PR #4, opened 2021-03-04
correctly returning instrument description dictionary

dsavchenko:


dsavchenko:
The output is still "raw" compared with the example in https://oda-api.readthedocs.io/en/latest/user_guide/TestAPI.html

volodymyrss:
ok! 

since you are onto it, could you please add a unit test for this?

volodymyrss:
> BTW, most of the tests are failing with connection error (to cdcihn.isdc.unige.ch/staging-1.2/dispatcher)

you can select only production-1-2 platform in parameterize: just add `-k production-1-2` to pytest. Also for the new tests.

And it's probably better to use that one by default in `get_platform_dispatcher`.

dsavchenko:
several last commits dedicated to errors from ci.yml workflow action

There were errors when running it, but checks passed regardless of it. I am not really into GitHub actions, but it seems to me, it is not the behaviour one should expect.

volodymyrss:
some errors are left intentionally in "xfail" approach.

It depends, which are the exact errors you are talking about?

dsavchenko:
I mean not the unit tests, but the action in .github/workflows/ci.yml
What is the reason to run such a long workflow in order to get an error inside and just ignore it?

volodymyrss:
> I mean not the unit tests, but the action in .github/workflows/ci.yml

Indeed, as you also said before. But the same approach apply.

> What is the reason to run such a long workflow in order to get an error inside and just ignore it?

To capture the error.
A lot of things can be improved.
A lot of them are not worth spending time to improve. 
A lot of them are not worth improving at this time. 
And sometimes the error will be made obsolete in the future.

First step to improve is to capture the error, and continue work on the more urgent things.

This is why xfail exists in pytest.
And why we consciously "forgive" some errors, knowing they exists, and coming back when/if needed.

dsavchenko:
Ok, I have understood the reasoning. 
What was weird to me is that errors in commands are ignored without any special annotation. Then I've realised that the multiline 'runs' are just bash-scripts, and one needs to return the exit status of the whole script explicitly. 
Now all is clear. Sorry for my ignorance.

volodymyrss:
> Ok, I have understood the reasoning.
> What was weird to me is that errors in commands are ignored without any special annotation. Then I've realised that the multiline 'runs' are just bash-scripts, and one needs to return the exit status of the whole script explicitly.

Some tests to fail when unexpected things happen.
The green check represents that.  And I think it's nice to have it as indicator of the minimal requirement.

But really a lot of tests in here and in some of the other places are skipped or ignored. Our approach to testing has not been always entirely uncontroversial.
We can see with time as to which tests should be elevated to this "critical" level.

> Now all is clear. Sorry for my ignorance.

No problem!
I appreciate, but I also I do not necessarily there think there is a need to apologize: as long as the question or comment is made in good faith it can generally only for the benefit all of us.

PR #31, opened 2021-05-26
Reasonable set api code

burnout87:


ferrigno:
In the last line, it would be useful to assign the products to a variable
e.g.
data_collection = disp.get_product(**par_dict)

ferrigno (approved):
The modifications you made are OK, but see my other two comments, if relevant.

oda_api/api.py
```diff
@@ -850,7 +850,7 @@ def set_api_code(query_dict):
         from oda_api.api import DispatcherAPI\n
         disp=DispatcherAPI(url='www.astro.unige.ch/cdci/astrooda/dispatch-data', instrument='mock', protocol='https')'''
```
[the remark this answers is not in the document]
volodymyrss: It is feasible, and can be passed as a parameter.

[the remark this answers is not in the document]
burnout87: We can have as a parameter in the dispatcher configuration and pass it to the function, as the dispatcher is the only place where this function is called, it should not be a problem. And the current value could be left as a default one.

oda_api/api.py
```diff
@@ -850,7 +850,7 @@ def set_api_code(query_dict):
...
 
-        _cmd_prod_ = 'disp.get_product(**par_dict)'
+        _cmd_prod_ = 'data_collection = disp.get_product(**par_dict)'
 
         _api_dict = {}
         for k in query_dict.keys():
```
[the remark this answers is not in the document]
burnout87: yes it should be

oda_api/api.py
```diff
@@ -866,7 +866,7 @@ def set_api_code(query_dict, url="www.astro.unige.ch/cdci/astrooda/dispatch-data
 
         _cmd_ = '%s\n' % _header
         _cmd_ += 'par_dict='
-        _cmd_ += '%s' % _api_dict
+        _cmd_ += '%s' % json.dumps(_api_dict, indent=0)
```
volodymyrss, line 869: why not put a conventional python indent? E.g. 4. And also the rest of the code could have formatting as conventional, spaces and all.

oda_api/api.py
```diff
@@ -837,7 +837,7 @@ def get_product(self,
         return d
 
     @staticmethod
-    def set_api_code(query_dict):
+    def set_api_code(query_dict, url="www.astro.unige.ch/cdci/astrooda/dispatch-data"):
```
volodymyrss, line 840: so how will it look like? With tesT?

volodymyrss (changes requested):

oda_api/api.py
burnout87: on it, but on a PR on the dispatcher-app

volodymyrss: could be here too with dispatcher-fixture. depends which code is tested. as you like

burnout87: actrually is faster

volodymyrss: ok!
in principle, would be good to also make API-focused tests with fixture here: https://github.com/oda-hub/oda_api/issues/13
but really after

tests/test_basic.py
```diff
@@ -0,0 +1,62 @@
...
+        if k in _alias_dict.keys():
+            n = _alias_dict[k]
+        else:
+            n = k
+        _api_dict[n] = param_dic[k]
+
+    expected_api_code = f'''from oda_api.api import DispatcherAPI
+disp=DispatcherAPI(url='http://0.0.0.0:8001', instrument='mock', protocol='https')
+
+par_dict={json.dumps(_api_dict, indent=4)}
```
volodymyrss, line 57: I would really spell all parameters out here, since we are trying to inspect how it looks like  :)
oh, and we could even do `eval` or something like that to check that the code is working. I did not realize that the one before used  cookies argument which is not useful for public at all and very confusing..

volodymyrss (changes requested):

tests/test_basic.py
burnout87: You mean just `url`? `instrument` and `protocol`, in that specific line are always set with that value and are not parametrized.

volodymyrss: I mean that entire expected_api_code can be just a string.
Right now, _api_dict is some code itself, repeating what is done in the API. So you are testing the code by repeating it in the test :)
If this code piece is somehow unfortunate, it would be likely still compatible with test, and not fail.

tests/test_basic.py
```diff
@@ -0,0 +1,62 @@
...
+        f_params = open(scratch_path_aliased + '/analysis_parameters.json')
+
+    jdata_output = json.load(f_query_output)
+
+    assert 'prod_dictionary' in jdata_output
+    assert 'api_code' in jdata_output['prod_dictionary']
+
+    output_api_code = jdata_output['prod_dictionary']['api_code']
+
+    param_dic = json.load(f_params)
```
volodymyrss, line 37: this till line 52 all a copy of the oda_api code, right? Can be removed and expectation spelled out, since you know what you expect.

burnout87: Ok, next commit makes the whole expected api output code explicit and not inferred/calculated. And it's lot less code :) Thanks for the suggestion

tests/test_basic.py
```diff
@@ -0,0 +1,50 @@
...
+        f_query_output = open(scratch_path_aliased + '/query_output.json')
+
+    jdata_output = json.load(f_query_output)
+
+    assert 'prod_dictionary' in jdata_output
+    assert 'api_code' in jdata_output['prod_dictionary']
+    output_api_code = jdata_output['prod_dictionary']['api_code']
+
+    expected_api_code = '''from oda_api.api import DispatcherAPI
+disp=DispatcherAPI(url='http://0.0.0.0:8001', instrument='mock', protocol='https')
```
volodymyrss, line 35: oh protocol is no more necessary. especially if it mismatches the url

burnout87: is it always provided within the url?

volodymyrss: it's not user, except it's validated. It should be left for compatibility, but it makes no sense, similarly to host.
URL by definintion contains protocol, host, port, path (as well as possibly credentials). The rest was some misunderstanding which eventually caused issues and was replaced.

volodymyrss (approved):

PR #35, opened 2021-05-26
Resolving several auth notebook problems: multiple request problem

volodymyrss:


volodymyrss:
a bit of a mix of small things.

oda_api/api.py
```diff
@@ -385,13 +404,13 @@ def parameters_dict_payload(self):
...
 
         if self.is_submitted:
-            return {
-                **p,
-                'job_id': self.job_id,
-                'query_status': self.query_status,
-            }
+             return {
+                 **p,
+        #         'job_id': self.job_id,
```
burnout87, line 409: An alternative could be to re-assign the `job_id` with the one returned from the dispatcher call, if any.I notived how the one returned from the dispatcher is not actually used

oda_api/api.py
```diff
@@ -491,7 +510,7 @@ def poll(self, verbose=None, silent=None):
         self.response_json = self.request_to_json()
         # <
             
-        logger.info("job: %s session %s", self.response_json['job_monitor']['session_id'], self.response_json['job_monitor']['job_id'])
+        logger.info("session: %s job: %s", self.response_json['job_monitor']['session_id'], self.response_json['job_monitor']['job_id'])
```
burnout87, line 513: good, this output was indeed confusing me :)

burnout87 (changes requested):
some thinking based on my still-limited knowledge of the oda_api

oda_api/api.py
burnout87: Perhaps here

[https://github.com/oda-hub/oda_api/blob/a124d8fd8e96858962a34495598df909e6c6cc5b/oda_api/api.py#L291-L342](https://github.com/oda-hub/oda_api/blob/a124d8fd8e96858962a34495598df909e6c6cc5b/oda_api/api.py#L291-L342)

At the completion of the request

burnout87: I think applying the change here would be ideal:

[https://github.com/oda-hub/oda_api/blob/a124d8fd8e96858962a34495598df909e6c6cc5b/oda_api/api.py#L507-L508](https://github.com/oda-hub/oda_api/blob/a124d8fd8e96858962a34495598df909e6c6cc5b/oda_api/api.py#L507-L508)

volodymyrss: How will reassigning the job_id after the request prevent sending incorrect, previously used, job_id in the next call?

volodymyrss: I will however add a check on returned job_id, in case it is different from the currently recorded.
If we want to keep sending job_id sometimes, we have to reset it when new parameters are set.  But it's difficult to know for sure when they changed, they might have been replaced with identical ones.

burnout87: Because that will be resent with the following request.

> I will however add a check on returned job_id, in case it is different from the currently recorded.
Exactly, that was indeed the check that I had in mind.
> If we want to keep sending job_id sometimes, we have to reset it when new parameters are set. But it's difficult to know for sure when they changed, they might have been replaced with identical ones.
> 
> http://www.faqs.org/rfcs/rfc1738.html

That is true, but the `job_id` is generated without considering the token as well as other parameters. And, in case a token is provided, the username is used.

volodymyrss: > Because that will be resent with the following request.

But it will be reset already after the faulty request.

> 
> > I will however add a check on returned job_id, in case it is different from the currently recorded.
> 
> Exactly, that was indeed the check that I had in mind.

This just a safety check, it will either do nothing or fail.

> 
> > If we want to keep sending job_id sometimes, we have to reset it when new parameters are set. But it's difficult to know for sure when they changed, they might have been replaced with identical ones.
> 
> That is true, but the `job_id` is generated without considering the token as well as other parameters. And, in case a token is provided, the username is used.

indeed, this makes it even worse. oda_api can not reasonably construct job_id except by calling dispatcher. Which is why it should not assume to derive it.

tests/test_live.py
```diff
@@ -299,3 +303,76 @@ def test_peculiar_request_causing_pickling_problem(platform):
...
+                osa_version="OSA10.2",
+                E1_keV=25.0,
+                E2_keV=80.0,
+                max_pointings=1000,
+            )
+
+
+def test_reusing_disp_instance(dispatcher_live_fixture):
+    from oda_api.api import RequestNotUnderstood 
+
```
burnout87, line 328: I guess this is not needed within this test

burnout87 (approved):

PR #39, opened 2021-05-27
check for None value set api_code and test alignement for url

burnout87:


burnout87:
As discussed, it is a bit tricky to reproduce as a behavior and therefore to write a test

oda_api/api.py
```diff
@@ -888,7 +888,8 @@ def set_api_code(query_dict, url="www.astro.unige.ch/cdci/astrooda/dispatch-data
                 else:
                     n = k
 
-                _api_dict[n] = query_dict[k]
+                if query_dict[k]:
```
volodymyrss, line 891: this might or might not be `None`. Could be `0`. Or False. Both may happen in parameters.
We are only concerned here with `None` to `null` morphing.

just `query_dict[k] is None` would be correct.

In principle, json is still not guaranteed to represent python code.
It would work for a subset of types only. That can be verified and error thrown if not satisfactory.

volodymyrss (changes requested):

oda_api/api.py
volodymyrss: I meant `query_dict[k] is not None`, sorry.

burnout87:
@volodymyrss  Can I merge this?

volodymyrss (approved):

PR #40, opened 2021-05-27
Token expired 403 more info

volodymyrss:


tests/test_basic.py
```diff
@@ -32,7 +32,7 @@ def test_oda_api_code(dispatcher_live_fixture):
     output_api_code = jdata_output['prod_dictionary']['api_code']
 
     expected_api_code = f'''from oda_api.api import DispatcherAPI
-disp=DispatcherAPI(url='http://0.0.0.0:8001', instrument='mock')
+disp=DispatcherAPI(url='http://www.astro.unige.ch/cdci/astrooda_', instrument='mock')
```
burnout87, line 35: I had already changed here with a previous PR, that probably was missed

burnout87 (approved):

tests/test_basic.py
burnout87: #39 is the one

Issue #44, opened 2021-06-07
inconsistency in ISGRI and JEM-X product show


ferrigno:
from oda_api.api import DispatcherAPI
disp=DispatcherAPI(url='http://dispatcher.staging.internal.odahub.io', instrument='mock')
par_dict={\
"src_name": "4U 1700-377",\
"RA": "270.80",\
"DEC": "-29.80",\
"T1": "2021-05-01",\
"T2": "2021-05-05",\
"T_format": "isot",\
"instrument": "jemx",\
"osa_version": "OSA11.0",\
"radius": "4",\
"max_pointings": "50",\
"integral_data_rights": "all-private",\
"jemx_num": "1",\
"E1_keV": "3",\
"E2_keV": "20",\
"product_type": "Real",\
"detection_threshold": "5",\
"product": "jemx_lc",\
"time_bin": "4",\
"time_bin_format": "sec",\
"catalog_selected_objects": "1,2,3",
"selected_catalog" :'{"cat_frame": "fk5", "cat_coord_units": "deg", "cat_column_list": [[0, 1, 2], ["GX 5-1", "MAXI SRC", "H 1820-303"], [96.1907958984375, 74.80066680908203, 66.31670379638672], [270.2771301269531, 270.7560729980469, 275.914794921875], [-25.088342666625977, -29.84027099609375, -30.366628646850586], [0, 1, 0], [0.05000000074505806, 0.05000000074505806, 0.05000000074505806]], "cat_column_names": ["meta_ID", "src_names", "significance", "ra", "dec", "FLAG", "ERR_RAD"], "cat_column_descr": [["meta_ID", "<i8"], ["src_names", "<U10"], ["significance", "<f8"], ["ra", "<f8"], ["dec", "<f8"], ["FLAG", "<i8"], ["ERR_RAD", "<f8"]], "cat_lat_name": "dec", "cat_lon_name": "ra"}',
[...]

PR #49, opened 2021-06-18
Show output formatting

burnout87:


burnout87:
Output of a Dummy request for `isgri_lc` is:

`ID=0 prod_name=isgri_lc_0_OAO1657m415  meta_data: {'src_name': 'OAO 1657-415', 'time_bin': 0.0115739687598762, 'time': 'TIME', 'rate': 'RATE', 'rate_err': 'ERROR'}`

oda_api/api.py
```diff
@@ -930,10 +930,11 @@ def __init__(self, data_list, add_meta_to_name=['src_name', 'product'], instrume
 
             name, var_name = self._build_prod_name(
                 data, name, add_meta_to_name)
-            setattr(self, var_name, data)
+            # TODO what is the reason for this ?
+            # setattr(self, var_name, data)
```
volodymyrss, line 934: This makes the data available as attributes of the object.
One may argue how convenient it is, surely it is sometimes and sometimes it causes issues.

But, as of now, it **must remain** since it is interface used in various notebooks, some of them published.

tests/test_basic.py
```diff
@@ -1,5 +1,46 @@
...
+                                    product_type="Dummy",
+                                    osa_version='OSA11.0',
+                                    token=encoded_token
+                                    )
+
+    products.show()
+
+    captured = capsys.readouterr()
+    print("Show() output: ", captured.out)
+    assert captured.out == "ID=0 prod_name=isgri_lc_0_OAO1657m415  meta_data: {'src_name': 'OAO 1657-415', 'time_bin': 0.0115739687598762, 'time': 'TIME', 'rate': 'RATE', 'rate_err': 'ERROR'}\n\n"
```
volodymyrss, line 43: Can we also here check that the result is available as `products.isgri_lc_0_OAO1657m415` and also in products._n_list and products._p_list

volodymyrss (changes requested):

volodymyrss (approved):
For me it's ok. I do not know if that's what @ferrigno wanted, I did not look into the issue too closely.

burnout87:
> For me it's ok. I do not know if that's what @ferrigno wanted, I did not look into the issue too closely.

Waiting for his approval indeed

volodymyrss:
> > For me it's ok. I do not know if that's what @ferrigno wanted, I did not look into the issue too closely.
> 
> Waiting for his approval indeed

For @ferrigno it could be useful to look into the [test](https://github.com/oda-hub/oda_api/blob/show-output-formatting/tests/test_basic.py#L43) to understand if that's what you expected to happen.

burnout87:
> For me it's ok. I do not know if that's what @ferrigno wanted, I did not look into the issue too closely.

Ah BTW, about my comment here:

[https://github.com/oda-hub/oda_api/issues/44#issuecomment-863867220](https://github.com/oda-hub/oda_api/issues/44#issuecomment-863867220)

Could this be an issue on the JEM-X side?

volodymyrss:
> > For me it's ok. I do not know if that's what @ferrigno wanted, I did not look into the issue too closely.
> 
> Ah BTW, about my comment here:
> 
> [#44 (comment)](https://github.com/oda-hub/oda_api/issues/44#issuecomment-863867220)
> 
> Could this be an issue on the JEM-X side?

Well that's a property, do not know if it is an issue, the way the source names are extracted they are guaranteed to squash some characters:

https://github.com/oda-hub/dispatcher-plugin-integral/blob/master/cdci_osa_plugin/osa_lightcurve_query.py#L174

It's confusing the variable is called source name. Actually source name is stored in the file and can be extracted if necessary.
The concern here in this issue though is just mismatch between printing and assigning.

Anyway these are not big issues.

ferrigno:
Source name should be derived as for ISGRI 
https://github.com/oda-hub/dispatcher-plugin-integral/blob/47c362111dfdee32c9d46fc5d15ae666af4ab0cb/cdci_osa_plugin/osa_lightcurve_query.py#L142

The way in which it is derived now for JEM-X
https://github.com/oda-hub/dispatcher-plugin-integral/blob/master/cdci_osa_plugin/osa_lightcurve_query.py#L174
leads to use a space instead of Minus or plus

So it is not exactly what I wanted for JEM-X.

ferrigno:
If you implement the change, do not transfer the "TIMEDEL" stuff to JEM-X, it will not work.

volodymyrss:
Indeed:

> Well that's a property, do not know if it is an issue, the way the source names are extracted they are guaranteed to squash some characters:
> 
> https://github.com/oda-hub/dispatcher-plugin-integral/blob/master/cdci_osa_plugin/osa_lightcurve_query.py#L174
> 
> It's confusing the variable is called source name. Actually source name is stored in the file and can be extracted if necessary.
> The concern here in this issue though is just mismatch between printing and assigning.
> 
> Anyway these are not big issues.

But it is a different issue, and belongs to dispatcher plugin.

This PR one here presumably had something to do with that what is printed is not what is available. Is this not what you explained?

The fact that what is provided by jemx is not unmodified src names can not be addressed by oda_api PR.

burnout87:
> Indeed:
> 
> > Well that's a property, do not know if it is an issue, the way the source names are extracted they are guaranteed to squash some characters:
> > https://github.com/oda-hub/dispatcher-plugin-integral/blob/master/cdci_osa_plugin/osa_lightcurve_query.py#L174
> > It's confusing the variable is called source name. Actually source name is stored in the file and can be extracted if necessary.
> > The concern here in this issue though is just mismatch between printing and assigning.
> > Anyway these are not big issues.
> 
> But it is a different issue, and belongs to dispatcher plugin.
> 
> This PR one here presumably had something to do with that what is printed is not what is available. Is this not what you explained?
> 
> The fact that what is provided by jemx is not unmodified src names can not be addressed by oda_api PR.

That was indeed my point. If we want to change the behavior, specifically for JEM-X then we'll do so on  a dedicated PR

ferrigno (approved):

ferrigno:
Right, I approve this one and close it

ferrigno:
I should have not closed it, I let you merge and close.

PR #50, opened 2021-06-18
removed one column

burnout87:
just noticed one, perhaps not needed, column


burnout87:
Update, just noticed that all notebooks are failing, so perhaps this is ok, I will close it

PR #55, opened 2021-07-20
testing and failing json encoding of vlf fits_rec

volodymyrss:
it is caused by [problem in astropy](https://github.com/astropy/astropy/pull/11957), encoding VLF column in FITS_rec . Jsonification in this case silently produces something quite different.
It happens currently only for OSA which uses newer heasoft, with `ftrbnrmf` always producing compressed rmf.

### DIFF
--- a/oda_api/api.py
+++ b/oda_api/api.py
@@ -882,7 +882,7 @@ class DispatcherAPI:
         return d
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


burnout87 (approved):
It looks ok, but I am not quite aware of the context of the related issue within astropy

PR #56, opened 2021-07-23
basic token cli: inspect

volodymyrss:


volodymyrss:
> Will the functionalities implemented here for the modification of the token cover what we discussed here [oda-hub/dispatcher-app#188](https://github.com/oda-hub/dispatcher-app/issues/188) ?

Partially. This is mostly for user-side manipulations. 
Normally, user will not have the secret key, so the modification manipulation will not work for them.
So endpoint of dispatcher is still needed. 
In fact, this change here should be coordinated with this new endpoint to use it instead of local secret key.
The interface for the user can remain.

burnout87:
> > Will the functionalities implemented here for the modification of the token cover what we discussed here [oda-hub/dispatcher-app#188](https://github.com/oda-hub/dispatcher-app/issues/188) ?
> 
> Partially. This is mostly for user-side manipulations.
> Normally, user will not have the secret key, so the modification manipulation will not work for them.
> So endpoint of dispatcher is still needed.
> In fact, this change here should be coordinated with this new endpoint to use it instead of local secret key.
> The interface for the user can remain.

I am indeed working on it and I was actually planning on using one of the function you implemented ( [https://github.com/oda-hub/oda_api/blob/89af64cadebc034ddd65e4ce366d87ce6273ea80/oda_api/token.py#L111](https://github.com/oda-hub/oda_api/blob/89af64cadebc034ddd65e4ce366d87ce6273ea80/oda_api/token.py#L111) ) once complete.

volodymyrss:
> > > Will the functionalities implemented here for the modification of the token cover what we discussed here [oda-hub/dispatcher-app#188](https://github.com/oda-hub/dispatcher-app/issues/188) ?
> > 
> > 
> > Partially. This is mostly for user-side manipulations.
> > Normally, user will not have the secret key, so the modification manipulation will not work for them.
> > So endpoint of dispatcher is still needed.
> > In fact, this change here should be coordinated with this new endpoint to use it instead of local secret key.
> > The interface for the user can remain.
> 
> I am indeed working on it and I was actually planning on using one of the function you implemented (
> 
> https://github.com/oda-hub/oda_api/blob/89af64cadebc034ddd65e4ce366d87ce6273ea80/oda_api/token.py#L111
> ) once complete.

sounds good!

oda_api/api.py
```diff
@@ -154,6 +157,11 @@ def func_wrapper(*args, **kwargs):
 
 
 class DispatcherAPI:
+
+    # allowing token discovery by default changes the user interface in some cases, 
+    # but in desirable way
+    token_discovery_methods = None
+
```
burnout87, line 164: The changes you talk about are related to the fact that the token might issued even for cases where it should not?

oda_api/token.py
```diff
@@ -0,0 +1,140 @@
...
+def decode_oauth2_token(token: str):
+    # usually comes in cookies['_oauth2_proxy']
+    return json.loads(base64.b64decode(token.split(".")[0]+"=").decode())
+
+## preserving
+
+def discover_token(token_discovery_methods=(
+        "environment variable ODA_TOKEN",
+        "file in current directory",
+        "file in home")):
```
burnout87, line 72: Are those ways to identify a token-discovery method? Perhaps a more compact naming would make it more easy to maintain ?

burnout87 (changes requested):
I really like this approach for the token discovery, I would have a suggestion to make it more robust

oda_api/api.py
volodymyrss: not at all. This would not be desirable.

What is implemented is, in fact, desirable.

Also, this part is not related to generating token. 
Only to discovering token.
The change is that now, if there is a discoverable (in env variable, or a file) token, it will be used by default. Before, token was only used if explicitly provided.

So now it is possible to preserve user own code with no explicit reference to tokens, and provide token at runtime. Less work for user, and less chance to embed token in shared code.

oda_api/token.py
volodymyrss: It's also use in output, so it's more informative.
Else we'd need informative name and a compact one.
Some sort of more controlled set of discovery methods could be used in the future.
Actually, I would switch to something like `dynaconf`, which does this internally.

oda_api/api.py
burnout87: Ok, that's indeed what I was thinking, and it makes sense.

burnout87 (approved):

Issue #59, opened 2021-08-05
unwanted warning on oda_api code


ferrigno:
I got this api_code from the frontend (very nice feature!) but I get a slightly confusing warning

```python
from oda_api.api import DispatcherAPI
disp=DispatcherAPI(url='https://www.astro.unige.ch/mmoda//dispatch-data', instrument='mock')

par_dict={
    "src_name": "GRB170325B",
    "RA": "",
    "DEC": "",
    "T1": "2017-03-25T21:49:01.00",
    "T2": "2017-03-25T21:51:01.00",
    "T_format": "isot",
    "instrument": "spi_acs",
    "product_type": "Real",
    "product": "spi_acs_lc",
    "time_bin": "1"
}
data_collection = disp.get_product(**par_dict)
```

(with logging at INFO level gives )

```
unable to dig list, instance not a dict by <class 'NoneType'>; object was None
/home/ferrigno/Soft/myVE-py37/lib/python3.7/site-packages/oda_api/api.py:846: UserWarning: 
----------------------------------------------------------------------------
the parameter: T_format   is not among valid ones:
['src_name', 'RA', 'DEC', 'T1', 'T2', 'token', None, 'selected_catalog', 'time_bin']
this will throw an error in a future version 
and might breack the current request!
 ----------------------------------------------------------------------------
```

volodymyrss:
It's sufficient to make a one-line change (warning => debug). But maybe it's useful if @burnout87 looks on that part to be familiar why this warning is even there. What do you think, @burnout87 ?

burnout87:
Ok, I am actually doing some debugging, but it seems that it comes from the way the spi_acs instrument is initialized.

burnout87:
Ok, I was wrong, it has nothing to do with the way the plugin of `spi_acs` is initialized.

It is rather because `T_format` is not actually treated as a parameter, but rather used in different context, which I haven't understood yet. 

But since it appears in the list of input parameters the warning is issued.

I will change to a debug level log message.

volodymyrss:
Removed with #61 unless  not.

burnout87:
Apparently the message still appears, as noted by @ferrigno

volodymyrss:
and still appears?

burnout87:
I remember that probably @ferrigno  was missing the most up -to-date version of `oda_api`. Did you double check @ferrigno ?

ferrigno:
I have not seen unwanted warning lately.

PR #61, opened 2021-08-05
Debugging level logging list inspection

burnout87:


oda_api/api.py
```diff
@@ -816,7 +816,7 @@ def get_product(self,
             instrument=instrument, product_type=product), cookies=self.cookies)
 
         if res.status_code != 200:
-            warnings.warn(
```
volodymyrss, line 819: although we did not use `warnings` much, it's not the same as `logging.warning`. See [here](https://docs.python.org/3/howto/logging.html). 

```
`warnings.warn()` in library code if the issue is avoidable and the client application should be modified to eliminate the warning

`logging.warning()` if there is nothing the client application can do about the situation, but the event should still be noted
```

In this case, we actually want the client to do something: adapt parameters. So, following the python convention, it should be `warnings.warn`

oda_api/api.py
```diff
@@ -841,9 +841,9 @@ def get_product(self,
                         msg += '%s' % valid_names
                         msg += '\n'
                         msg += 'this will throw an error in a future version \n'
-                        msg += 'and might breack the current request!\n '
```
volodymyrss, line 844: I think it's `break` here.

oda_api/api.py
```diff
@@ -841,9 +841,9 @@ def get_product(self,
...
+                        msg += 'and might brake the current request!\n '
                         msg += '----------------------------------------------------------------------------\n'
-                        warnings.warn(msg)
```
volodymyrss, line 846: this is actually a warning. And, following note above, not `logging.warn` bit `warnings.warn`

volodymyrss (changes requested):

oda_api/api.py
burnout87: According to what we discussed in #59 , I thought this was supposed to be changed to debug level message. Or perhaps I misunderstood.

volodymyrss: > According to what we discussed in #59 , I thought this was supposed to be changed to a debug level message. Or perhaps I misunderstood.

The part that lead to `unable to dig list, instance not a dict by...` should be.  This is utterly incomprehensible for the user.

`invalid parameter` warning should remain. You can imagine it is important that user is warned (at least!) when using incorrect parameter name.
It should be separately determined why, as you noted as well, `T_format` is not understood. It should be, in principle, understood and used correctly. Maybe not in all queries though.
I guess it was passed by frontend API.
In any case, at this level, it should be clear to user that this parameter is not used.

burnout87: Ok, now is clear. I was confused and thought we were referring to the warning message that should have been changed.

As of `T_format`, I had indeed the same impression.

volodymyrss (approved):
thanks! please rename though.

PR #72, opened 2021-08-31
Host url compatibility

burnout87:


oda_api/api.py
```diff
@@ -191,13 +188,20 @@ def __init__(self,
             if host.startswith('http'):
                 self.url = host
             else:
-                if protocol == 'http':
-                    self.url = "http://" + host
-                elif protocol == 'https':
-                    self.url = "https://" + host
-                else:
+                if protocol != 'http' and protocol != 'https':
```
volodymyrss, line 191: it is, surely, better, but we should really progress to discard this part, not improve :)

oda_api/api.py
```diff
@@ -191,13 +188,20 @@ def __init__(self,
...
+                    self.url = protocol + "://" + host
         else:
+            if url is None:
+                url = "https://www.astro.unige.ch/mmoda/dispatch-data"
+            else:
+                if not url.startswith('http'):
+                    if protocol != 'http' and protocol != 'https':
+                        raise UserError('protocol must be either http or https')
+                    else:
+                        url = protocol + "://" + url
```
volodymyrss, line 203: this is very incorrect, URL already includes "protocol" (actually, scheme, which represents protocol in subtle way), by [definition](https://datatracker.ietf.org/doc/html/rfc3986). If `url` is provided,  every other connection parameter (port, host, protocol) should be discarded. And later obsolete.

The fact that this division of URL was even introduced is a result of misunderstanding of what URL is.

volodymyrss (changes requested):

oda_api/api.py
```diff
@@ -175,10 +175,7 @@ def __init__(self,
                  wait=True,
                  n_max_tries=20,
                  session_id=None,
-                 ):
-
-        if url is None:
-            url = "https://www.astro.unige.ch/mmoda/dispatch-data"
+                 ) -> object:
```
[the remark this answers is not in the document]
burnout87: Absolutely! This must be the result of an erroneous copy and past.

oda_api/api.py
burnout87: Since I didn't know the real reason why `host` and `url` were both still available as parameter, I didn't want to, for the moment, change too much. 

But sure, it has to be improved

burnout87: Do we want still want to have the `host` parameter available? Is it still necessary?

burnout87: > The warning is already given, see there. Next step to discard. I think we could even make a more severe warning. It costs little to maintain current behavior.

So I will leave for now the `host` parameter compatibility.

> 
> But the url case should be implemented well, as it's the good way.

What is missing  to it?

burnout87: now a deprecation warning is sent to the user, with a recommendation to change the implementation

oda_api/misc_helpers.py
```diff
@@ -0,0 +1,31 @@
+import re
+from oda_api.api import UserError
+
+
+regex_url = re.compile(
+                r'^(?:http|ftp)s?://'  # http:// or https://
```
volodymyrss, line 4: no ftp

oda_api/api.py
```diff
@@ -181,23 +180,27 @@
...
-                elif protocol == 'https':
-                    self.url = "https://" + host
-                else:
+                if protocol != 'http' and protocol != 'https':
                     raise UserError('protocol must be either http or https')
+                else:
+                    self.url = protocol + "://" + host
         else:
+            if not oda_api.misc_helpers.validate_url(url):
+                raise UserError(f'{url} is not a valid url, '
```
volodymyrss, line 201: maybe let's add here a more detailed explanation of what might be wrong. E.g. suggest user to add schema to the url, etc. Even just as a comment. Or even give an example of the url? Like, when requests.get gets wrong URL is suggests "did you mean this: ...". Just a comment would be enough.

oda_api/api.py
```diff
@@ -181,23 +180,27 @@
...
+                                f'please check it and try to issue again the request')
```
volodymyrss: ok, but why not use https://www.astro.unige.ch/mmoda/.....? :)

oda_api/api.py
```diff
@@ -181,23 +180,27 @@
...
+
```
volodymyrss, line 205: people do not know what schema is, let's also write "...schema, i.e. url should start with http:// or https:// "

oda_api/api.py
```diff
@@ -181,23 +180,30 @@
...
-                elif protocol == 'https':
-                    self.url = "https://" + host
-                else:
+                if protocol != 'http' and protocol != 'https':
                     raise UserError('protocol must be either http or https')
+                else:
+                    self.url = protocol + "://" + host
         else:
+            if not oda_api.misc_helpers.validate_url(url):
+                raise UserError(f'{url} is not a valid url. \n'
```
volodymyrss, line 201: could you put actually usable url? try with this one, it will not work :)

volodymyrss (changes requested):
> are you talking about the example url ? because here I see the older version of the code 

yes. Example URL should be usable. I mean, it's not critical, but it's likely to trigger errors, why not put a usable url?..

oda_api/api.py
burnout87: you mean: `{url} is not a valid url` -> `{url} is not a usable url` ?

volodymyrss: Oh no. Sorry comment went to the wrong line, I meant it for the block.
I mean example `astro.unige...` url you put in the suggestion is not a usable URL for oda api. Sorry in my comment I put dots at the end, i meant it has to be completed. It has dispatch-data at the end. It's could be just the real default url. Just try to use in oda api the url we give user as example.

burnout87: Perhaps is a good idea to have it indeed as a default url

tests/test_basic.py
```diff
@@ -102,3 +103,34 @@ def test_oda_api_code(dispatcher_api):
...
     assert output_api_code == expected_api_code or output_api_code == expected_api_code.replace('PRODUCTS_URL', 'http://0.0.0.0:8001')
+
+
+@pytest.mark.parametrize("protocol_url", ['http://', 'https://', ''])
+@pytest.mark.parametrize("init_parameter", ['host', 'url'])
+@pytest.mark.parametrize("protocol_parameter_value", ['http', 'https', '', None, 'not_included'])
+def test_host_url_init(dispatcher_live_fixture, protocol_url, init_parameter, protocol_parameter_value):
+    from oda_api.api import UserError
+
+    dispatcher_live_fixture_parameter = re.sub(r"https?://?", protocol_url, dispatcher_live_fixture)
```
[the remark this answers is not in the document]
burnout87: Oh yes, my bad, didn't actually see the mistake

oda_api/misc_helpers.py
```diff
@@ -0,0 +1,29 @@
+import re
+
+regex_url = re.compile(
+                r'^(?:http)s?://'  # http:// or https://
```
volodymyrss, line 4: you do not need the group here since there is no more ftp

volodymyrss (approved):

volodymyrss (approved):

PR #83, opened 2021-09-13
default url with None

volodymyrss:
this is easier to use because when I want to propagate a URL from elsewhere, I can propagate None and not add a condition on passing or not to `DispatcherAPI` constructor.

### DIFF
--- a/oda_api/api.py
+++ b/oda_api/api.py
@@ -169,7 +169,7 @@ class DispatcherAPI:
 
     def __init__(self,
                  instrument='mock',
-                 url="https://www.astro.unige.ch/mmoda/dispatch-data",
+                 url=None,
                  run_analysis_handle='run_analysis',
                  host=None,
                  port=None,
@@ -179,6 +179,9 @@ class DispatcherAPI:
                  n_max_tries=20,
                  session_id=None,
                  ):
+
+        if url is None:
+            url = "https://www.astro.unige.ch/mmoda/dispatch-data"
 
         if host is not None:
             msg = '\n'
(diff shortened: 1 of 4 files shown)


burnout87:
> this is easier to use because when I want to propagate a URL from elsewhere, I can propagate None and not add a condition on passing or not to `DispatcherAPI` constructor.

Are you referring to a case where you want  to use a different URL?

volodymyrss:
> > this is easier to use because when I want to propagate a URL from elsewhere, I can propagate None and not add a condition on passing or not to `DispatcherAPI` constructor.
> 
> Are you referring to a case where you want to use a different URL?

I am referring to the case when I want to propagate a URL from elsewhere. And want to use the default.

For example, nested call to constructor:

```python
def useful_function(dispatcher_url):
    DispatcherAPI(url=dispatcher_url)
```

how do I call it with default value? I have to set another default for useful_function!
Which means `useful_function` also needs to know about what is the exact default of oda_api. Which is repetition.

If I want to actually use the default and not repeat it in `useful_function` I have to:

```python
def useful_function(dispatcher_url=None):
    if url is None:
        DispatcherAPI()
    else:
        DispatcherAPI(url=dispatcher_url)
```

it can be put in one line with some dictionary kw, but it's not nice.

With the proposed change the original function works if I pass None.
The `None` for default is repeated everywhere, but it's commonly considered as reasonable ["unset value" default](https://realpython.com/python-optional-arguments/). This is a very common pattern, and it is used throughout the software we use (matplotlib, astropy,..).

burnout87 (approved):
Ok, thanks for the improvement.

Issue #84, opened 2021-09-14
error in oda_api v 1.1.21


ferrigno:
The API code received in the email gives a type error
TypeError: list indices must be integers or slices, not str

see below the call. **Note that I deleted my token, so please insert yours to reproduce it.**

```
from oda_api.api import DispatcherAPI

disp = DispatcherAPI(url="https://www.astro.unige.ch/mmoda//dispatch-data", instrument="mock")

par_dict = {
"osa_version": "OSA11.0-dev210827.0528-37487",
"E1_keV": "25.0",
"E2_keV": "80.0",
"integral_data_rights": "all-private",
"scw_list": [
"101900100010.001",
"101900130010.001",
"101900160010.001",
"101900190010.001",
"101900340010.001",
"101900370010.001",
"101900400010.001",
"101900540010.001",
"101900550010.001",
"101900560010.001",
"101900570010.001",
"101900580010.001",
"101900590010.001",
"101900600010.001",
"101900610010.001",
"101900620010.001",
"101900630010.001",
"101900640010.001",
"101900650010.001",
"101900660010.001",
"101900670010.001",
"101900680010.001",
"101900690010.001",
"101900700010.001",
"101900710010.001",
"101900720010.001",
"101900730010.001",
"101900740010.001",
"101900750010.001",
"101900760010.001",
"101900770010.001",
"101900790010.001",
"101900810010.001",
"101900830010.001",
"101900840010.001",
"101900860010.001",
"101900870010.001",
"101900880010.001",
"101900890010.001",
"101900900010.001",
"101900910010.001",
"101900920010.001",
"101900930010.001",
"101900940010.001",
"101900950010.001",
"101900960010.001",
"101900970010.001",
"101900980010.001",
"101901000010.001",
"101901010010.001",
"101901020010.001",
"101901030010.001",
"101901040010.001",
"101901050010.001",
"101901060010.001",
"101901070010.001",
"101901080010.001",
"101901090010.001",
"101901100010.001",
],
"selected_catalog": (
'{"cat_frame": "fk5", "cat_coord_units": "deg", "cat_column_list": [[0, 3], ["1A 0535+262",'
' "Crab"], [779.8522338867188, 1602.0269775390625], [84.72511291503906, 83.63396453857422],'
" [26.31719970703125, 22.01705551147461], [-32768, -32768], [2, 2], [0, 0],"
' [0.0002800000074785203, 0.0002800000074785203]], "cat_column_names": ["meta_ID",'
[...]

PR #85, opened 2021-09-14
Remove support dry run parameter

burnout87:
Any mention of the `dry_run` is removed.

Ideally we could consider doing the same within the dispatcher

### DIFF
--- a/oda_api/api.py
+++ b/oda_api/api.py
@@ -419,6 +419,7 @@ class DispatcherAPI:
         mismatching_parameters = []
         for k in self.parameters_dict.keys():
             # these do not correspond to meaning
+            # TODO can it be removed from here too? This function to be used only within a test
             if k in ['query_status', 'off_line', 'verbose', 'dry_run']:                
                 continue
 
--- a/oda_api/api.py
+++ b/oda_api/api.py
@@ -419,8 +419,14 @@ class DispatcherAPI:
         mismatching_parameters = []
         for k in self.parameters_dict.keys():
             # these do not correspond to meaning
-            # TODO can it be removed from here too? This function to be used only within a test
-            if k in ['query_status', 'off_line', 'verbose', 'dry_run']:                
+            ''' 
+            The dry_run parameter is not actually considered within the oda_api,
+            but we keep it here for  consistency.
+            As discussed in: 
+            * https://github.com/oda-hub/oda_api/pull/85
+            * https://github.com/oda-hub/oda_api/issues/84
+            '''
+            if k in ['query_status', 'off_line', 'verbose', 'dry_run']:
                 continue
 
             returned = self.returned_analysis_parameters.get(k, None)
--- a/oda_api/api.py
+++ b/oda_api/api.py
@@ -852,7 +852,7 @@ class DispatcherAPI:
 
         if 'dry_run' in kwargs:
             warnings.warn('The dry_run parameter you included is not going to have any effect on the execution.\n'
-                          'However the oda_api will perform a check of the list opf valid parameters for your request.')
+                          'However the oda_api will perform a check of the list of valid parameters for your request.')
 
         res = requests.get("%s/api/par-names" % self.url, params=dict(
             instrument=instrument, product_type=product), cookies=self.cookies)
--- a/oda_api/api.py
+++ b/oda_api/api.py
@@ -853,6 +853,7 @@ class DispatcherAPI:
         if 'dry_run' in kwargs:
             warnings.warn('The dry_run parameter you included is not going to have any effect on the execution.\n'
                           'However the oda_api will perform a check of the list of valid parameters for your request.')
+            del kwargs['dry_run']
 
         res = requests.get("%s/api/par-names" % self.url, params=dict(
             instrument=instrument, product_type=product), cookies=self.cookies)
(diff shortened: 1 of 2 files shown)


oda_api/api.py
```diff
@@ -419,6 +419,7 @@ def returned_analysis_parameters_consistency(self):
         mismatching_parameters = []
         for k in self.parameters_dict.keys():
             # these do not correspond to meaning
+            # TODO can it be removed from here too? This function to be used only within a test
```
volodymyrss, line 422: maybe let's add comment to remove with reference to the relevant discussion.
And keep for now, why not. It's safe.

oda_api/api.py
```diff
@@ -903,18 +903,8 @@ def get_product(self,
...
-
-        data = None
-
-        if not dry_run:
-            d = DataCollection.from_response_json(
-                res_json, instrument, product)
-
-        else:
-            self._decode_res_json(
-                res.json()['products']['instrument_parameters'])
```
volodymyrss, line 916: you may deduce that this did not work since above `res.json()` was a  list

oda_api/api.py
```diff
@@ -903,18 +903,8 @@ def get_product(self,
             raise RuntimeError(
                 "not failed, not, but complete? programming error for client!")
 
-        # <
-
-        data = None
-
-        if not dry_run:
```
volodymyrss, line 910: this was `True` when `dry_run == "False"`, i.e. for all API code in emails..

volodymyrss:
We should also remove it from doc, but @ferrigno can do it in a separate PR?

volodymyrss (approved):

oda_api/api.py
burnout87: indeed it was the point where the crash related to #84 was happening

volodymyrss: I do not completely understand why we did not see it before. This should have happened with all of the email API codes. Though only when product is ready.

volodymyrss: Can we make a test to make sure this completes?..

burnout87: Sure. But, should we assume that perhaps, in some cases, `dry_run` is still passed as a parameter to the `oda_api` ?

burnout87: Because in that case, we have to make sure that it won't be included within the output api_code

volodymyrss: Ok, and add a warning that it does nothing, and the request will be checked when it is sent.

I think dry_run functionality - checking parameters was previously done by dispatcher. Now list of valid parameters is inspected by oda_api.

burnout87: Indeed, in the dispatcher simply lists the parameters:
[https://github.com/oda-hub/dispatcher-app/blob/c0dcd03461fd7e1bb7aa9591814f46bb11ab05b8/cdci_data_analysis/analysis/instrument.py#L227](url)

oda_api/api.py
```diff
@@ -843,7 +849,10 @@ def get_product(self,
         kwargs['query_status'] = 'new',
         kwargs['verbose'] = verbose,
         kwargs['session_id'] = self.session_id
-        kwargs['dry_run'] = dry_run,
+
+        if 'dry_run' in kwargs:
+            warnings.warn('The dry_run parameter you included is not going to have any effect on the execution.\n'
+                          'However the oda_api will perform a check of the list opf valid parameters for your request.')
```
volodymyrss, line 855:
```suggestion
                          'However the oda_api will perform a check of the list of valid parameters for your request.')
```

oda_api/api.py
```diff
@@ -853,9 +862,10 @@ def get_product(self,
                 'parameter check not available on remote server, check carefully parameters name')
         else:
             _ignore_list = ['instrument', 'product_type', 'query_type',
-                            'off_line', 'query_status', 'verbose', 'session_id', 'dry_run']
+                            'off_line', 'query_status', 'verbose', 'session_id']
```
volodymyrss, line 865: so now there will be two warnings? And it will be sent, and will actually affect the analysis since dispatcher still supports it?

volodymyrss (changes requested):

oda_api/api.py
burnout87: You mean one from here and one from the dispatcher?

volodymyrss: I mean in validation in the next lines.

burnout87: In any case, depends on what we want to do: we can still have it ignored by the oda_api, as it was before, with the new warning saying that the parameter has no effect, then we remove it from the dispatcher.

Otherwise, for the moment, we can warn the user that the support for this parameter will end soon, but we still support it for the moment.

I personally would leave the warning we introduced now and remove the support from the dispatcher.

volodymyrss: > In any case, depends on what we want to do: we can still have it ignored by the oda_api, as it was before, with the new warning saying that the parameter has no effect, then we remove it from the dispatcher.
> 
> Otherwise, for the moment, we can warn the user that the support for this parameter will end soon, but we still support it for the moment.
> 
> I personally would leave the warning we introduced now and remove the support from the dispatcher.
> 
> But of course, I have no experience regarding the usage of such a parameter.

You can just remove it from `kwargs` right after you warn about it not being used. 
This will prevent an additional warning in parameter validation in oda_api, and also avoid any confusion about what is done by the dispatcher.

burnout87: I agree, it is the smoothest approach.

burnout87:
Just to double-check/-confirm: we keep the support of the `dry_run` parameter in dispatcher.

volodymyrss (approved):

PR #88, opened 2021-09-27
adjusting test oda_api

burnout87:


tests/test_basic.py
```diff
@@ -96,7 +96,13 @@ def test_oda_api_code(dispatcher_api):
     "product_type": "Dummy",
     "off_line": "False",
     "api": "True",
-    "oda_api_version": "{oda_api.__version__}"
+    "oda_api_version": "{oda_api.__version__}",
```
volodymyrss, line 99: cool, this is actually very useful. Assuming these defaults with no notice anywhere is kind of wrong. We could even warn about the true request in API. But it's not that critical.

volodymyrss (approved):

PR #94, opened 2021-11-22
Better token discovery

volodymyrss:


volodymyrss:
had to integrate in #95  else it does not pass tests

PR #95, opened 2021-11-22
Warning for no log and fixes for broken tests related to token discovery

volodymyrss:


burnout87 (commented):
One question

oda_api/api.py
```diff
@@ -838,6 +838,11 @@ def get_product(self,
         submit query, wait (if allowed by self.wait), decode output when found
         """
 
+        self.logger.warning('please beware that by default, in a typical setup, oda_api will not output much. '
+                            'To learn how to increase the verbosity, please refer to the documentation: '
+                            'https://oda-api.readthedocs.io/en/latest/user_guide/ScienceWindowList.html?highlight=logging#Let\'s-get-some-logging'
+                            )
+
```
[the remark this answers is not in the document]
volodymyrss: Because constructor never produces a message user really expects.

Instead, it is during `get_product` that user wants to see the feedback. It's a long call, and it is when the messages may be, in some cases, useful.

Constructor may be called once way above. Some other log may happen in after.
And then get_product will be still empty.

The communication is expected at `get_product` and that's where the warning should happen.

burnout87 (dismissed):

burnout87:
Can I review? I see you're still pushing changes.

Perhaps also renaming would be useful, since it now integrates also #94

volodymyrss:
> Can I review? I see you're still pushing changes.

Yeah, I am just dealing with broken readthedocs. Docs need to be slightly updated for this, and there are important new versions of dependencies.
 
> Perhaps also renaming would be useful, since it now integrates also #94

Yes, I will also squash I guess. Too much small comments.

volodymyrss:
rtd passed, now I just remove some unused code

oda_api/cli.py
```diff
@@ -66,20 +66,24 @@ def get(obj, instrument, product, argument):
...
-def tokencli(obj, secret):
+def tokencli(obj, secret, allow_invalid):
     obj['secret_key'] = secret
+    obj['token'] = discover_token(allow_invalid=allow_invalid)
+    obj['decoded_token'] = decode_oda_token(obj['token'], secret_key=obj['secret_key'])
+    obj['allow_invalid'] = allow_invalid
+
+    if obj['token'] is None:
+        logger.warn("no token found!")
+        return
```
burnout87, line 79: Perhaps this should go a couple of lines above? before decoding I mean

burnout87 (changes requested):

oda_api/cli.py
volodymyrss: Well, `decode_oda_token` will raise if `token is None` so this is really useless. I remove! And also I add test for this.

burnout87 (approved):

PR #100, opened 2021-12-05
Safe retry

volodymyrss:


oda_api/api.py
```diff
@@ -179,7 +181,7 @@ def __init__(self,
                  cookies=None,
                  protocol="https",
                  wait=True,
-                 n_max_tries=20,
+                 n_max_tries=200,
```
burnout87, line 184: Why is it necessary to have so many re-tries? Don't we risk to have a very long waiting time?

oda_api/api.py
```diff
@@ -153,13 +153,15 @@ def func_wrapper(*args, **kwargs):
...
+                    logger.debug("problem in API call, %i tries left:\n%s\n sleeping %i seconds until retry",
                                    n_tries_left, message, retry_sleep_s)
+                    logger.warning("possibly temporary problem in API call, %i tries left, sleeping %i seconds until retry",
+                                   n_tries_left, retry_sleep_s)                                   
                     time.sleep(retry_sleep_s)
-            
-            raise RemoteException(
-                message=message
-            )
+                else:
```
burnout87, line 161: with this branch, not much changed I guess right? Or is there an additional purpose?

burnout87 (commented):
a couple of questions

oda_api/api.py
volodymyrss: this ensures the retry actually happens. It happned not quite as expected.

Oh, I should add a test for it.

volodymyrss: test added. 
Note that the change in these lines, moving the `raise`, is actually strongly impacting the retry behavior. It enables it. Before, it was kind of broken - apparently.

oda_api/api.py
volodymyrss: We do not simply risk having a very long waiting time, we expect it. 
It better than having user stare in frustration on an in-comprehensive exception!

If it is our fault, it is us who should stare on the exception (in sentry)

We need to choose what should happen in such cases:
* either break and expect user to correct  (retry later? nothing the user can typically do)
* retry and wait until condition resolves itself

Since we can really not expect the user to do anything, waiting (almost indefinitely) is usually the best.
In special cases, like in batch, the behavior might be adjusted, if the jobs should actually break. But most users who prefer not to look deep would prefer to wait.

volodymyrss:
I think generally conversations are best resolved by those who started them, so that it's clear that the intent of the question is addressed.

oda_api/api.py
burnout87: Yes, now, the more I look at it the more I see it. I never realized before.

Also, I would remove that line (135, `UnexpectedDispatcherStatusCode` case), that message is never used, unless we want to use it when raising an exception

burnout87 (approved):
Thanks for the explanation, just one comment.

burnout87 (approved):

volodymyrss:
forgot to bump version

PR #109, opened 2021-12-14
post simple empty data product

burnout87:


tests/test_cli.py
```diff
@@ -75,8 +75,8 @@ def test_get(dispatcher_live_fixture, caplog):
     runner = CliRunner()
     result = runner.invoke(cli.cli, ['-u', dispatcher_live_fixture, 'get', '-i', 'empty'], obj={})
     assert result.exit_code == 0
-
-    assert "'prod_dict': {'dummy': 'empty_parameters_dummy_query', 'failing': '" in caplog.text
+    # TODO not sure it is supposed to be this way
```
volodymyrss, line 78: this depends on what is installed in the python environment. Presumably what's here should be what is installed in Actions. But it may differ from what you have locally.

volodymyrss (changes requested):
Could you add the example notebook, in the `doc`?

volodymyrss:
> Could you add the example notebook, in the `doc`?

Ok! But we should make sure it works!

burnout87:
> > Could you add the example notebook, in the `doc`?
> 
> Ok! But we should make sure it works!

Now it;s there, I wanted to test it, could you please grant me the roles for posting over the gallery?

volodymyrss:
> > > Could you add the example notebook, in the `doc`?
> > 
> > 
> > Ok! But we should make sure it works!
> 
> Now it;s there, I wanted to test it, could you please grant me the roles for posting over the gallery?

You can use the same token as in production. Staging instance is safe enough to trust it with this secret.
I set you the role.

burnout87:
> 
> You can use the same token as in production.

I still get the same error also with this token

volodymyrss:
> > You can use the same token as in production.
> 
> I still get the same error also with this token

The same error as I got?

burnout87:
> > > You can use the same token as in production.
> > 
> > 
> > I still get the same error also with this token
> 
> The same error as I got?

Yes exactly the same

volodymyrss:
> > > > You can use the same token as in production.
> > > 
> > > 
> > > I still get the same error also with this token
> > 
> > 
> > The same error as I got?
> 
> Yes exactly the same

So token works?

I also get something very similar (but strangely, not the same) when running local dispatcher.

here's the event:

http://cdciweb02.isdc.unige.ch/mmoda/gallery/admin/reports/dblog/event/2517749

volodymyrss (dismissed):
See in the nb how I print link to explore. Something like that might be convenient in the API. But dict is also useful. So we can also see later.

volodymyrss (approved):

burnout87:
Is the `test-live` necessary in this case?

volodymyrss:
> Is the `test-live` necessary in this case?

It should not have broken. It should always pass, in principle. It's only skipped because it's too much to run it with every commit.
Maybe we can make it necessary before merge.

volodymyrss:
If you are certain the test did not break because of the current change, you can skip it. But it did pass before. Something with plot_tools, and you touched this module here, so perhaps this PR breaks it.

burnout87:
> 
> If you are certain the test did not break because of the current change, you can skip it. But it did pass before. Something with plot_tools, and you touched this module here, so perhaps this PR breaks it.

I was missing a couple of checks in the plot_tools module, preventing the test from passing

burnout87:
@volodymyrss and the test passes now

volodymyrss (dismissed):

burnout87:
Is there a way to decrease the codecov target? Currently it's preventing me from merging

volodymyrss (approved):

volodymyrss:
> Where can the codecov target be decreased? Currently it's preventing me from merging

see now

burnout87:
> > Where can the codecov target be decreased? Currently it's preventing me from merging
> 
> see now

what did you change?

volodymyrss:
> > > Where can the codecov target be decreased? Currently it's preventing me from merging
> > 
> > 
> > see now
> 
> what did you change?

 the codecov target !

PR #116, opened 2022-01-12
Enforce label

volodymyrss:
this will ensure that we know, in the future, which PR breaks the live test (if any, it can also break because of the remote server, but this is exceptional).

edit: I changed this to master. So that we check if it failed before #109 . In any case, it's suspicious that #109  changes the same module as the breaking test uses (plot_tools)


volodymyrss:
@burnout87 you see live test  passes here. So #109 broke it, and should fix it.

burnout87 (approved):

PR #118, opened 2022-01-16
GW

dsavchenko:
All crucial functionality is implemented. The only thing that is left is some convenient datatype (with plotting function etc) for the contours data.


dsavchenko:
@volodymyrss do you know, what's wrong with this prose linting action?

volodymyrss:
> @volodymyrss do you know, what's wrong with this prose linting action?

yes, you need to update the frozen version, it became incompatible, like that https://github.com/oda-hub/doc-handbook/pull/18/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3L12-R12

volodymyrss:
Just one more thing: could you please update the docs? Just an example notebook will be sufficient.

oda_api/data_products.py
```diff
@@ -610,7 +610,12 @@ def __init__(self,cat_dict,name='catalog'):
         if 'cat_coord_units' in cat_dict.keys():
             coord_units = cat_dict['cat_coord_units']
 
-        meta = {'FRAME': frame}
+        if 'cat_meta' in cat_dict.keys():
+            meta = cat_dict['cat_meta']
+        else:
+            meta = {}
```
burnout87, line 616: What is `cat_dict['cat_meta']` supposed to contain? Is it an dict entry? 

Cause otherwise I would use the same approach of the other if branches of this function, and init `meta` at the top of the function.

burnout87 (changes requested):
One question

oda_api/data_products.py
dsavchenko: Yes, this is a dict containing catalog table metadata. It is attached to the catalog table in the backend (6th cell of https://github.com/oda-hub/gw-backend/blob/master/nb/conesearch.ipynb) and passed as-is through the plugin https://github.com/oda-hub/dispatcher-plugin-gw/blob/142a37cf8ee0e5527aa15dccdbd51920a3451d6c/dispatcher_plugin_gw/products.py#L277

I think that this way is rather universal, we don't need to know, which fields are there.

dsavchenko:
> Just one more thing: could you please update the docs? Just an example notebook will be sufficient. 

I still have to do this

oda_api/data_products.py
burnout87: ok then I would just suggest the following to have a more consistent code:

```suggestion
        if 'cat_meta' in cat_dict.keys():
            cat_meta_entry = cat_dict['cat_meta']
            
        meta.update(cat_meta_entry)
```

and have `meta = {}` at the top of the method

burnout87 (changes requested):
What about moving the GW specific plot functionalities within the `plot_tools` module?

It will be also very useful for the production of images to upload over the product-gallery.

burnout87 (changes requested):
I think the plot_tools is not at the latest version, that includes also methods specific for the generation of images suitable for the product gallery. I would update the branch and add those methods here as well for the `OdaGWContours` class

setup.py
```diff
@@ -53,6 +53,10 @@
                 "pytest-xdist[psutil]",
                 "astroquery>=0.4.4",
             ],      
+          'gw': [
```
volodymyrss, line 56: we should add it in the doc - in the notebook - that this should be installed

volodymyrss:
One extra PR for the doc, one comment. And also in the notebook - please put the production reference.

volodymyrss (changes requested):
small changes

dsavchenko:
> I think the plot_tools is not at the latest version, that includes also methods specific for the generation of images suitable for the product gallery. I would update the branch and add those methods here as well for the `OdaGWContours` class

Do you need some action from me in this PR?

volodymyrss:
> > I think the plot_tools is not at the latest version, that includes also methods specific for the generation of images suitable for the product gallery. I would update the branch and add those methods here as well for the `OdaGWContours` class
> 
> Do you need some action from me in this PR?

Not really. Just re-request the review (I did it).
Also consider that master was updated: please rebase/merge.

volodymyrss (approved):

burnout87:
> > I think the plot_tools is not at the latest version, that includes also methods specific for the generation of images suitable for the product gallery. I would update the branch and add those methods here as well for the `OdaGWContours` class
> 
> Do you need some action from me in this PR?

I was referring to the methods to generate the image for being then uploaded over the gallery

volodymyrss (approved):

PR #132, opened 2022-03-09
Adding new source func

burnout87:


tests/test_live.py
```diff
@@ -87,7 +87,7 @@ def test_instruments():
         host=get_platform_dispatcher(),
         instrument="mock",
     )
-    assert disp.get_instruments_list() == ['magic', 'isgri', 'jemx', 'polar', 'antares', 'spi_acs']
+    assert disp.get_instruments_list() == ['isgri', 'jemx', 'polar', 'antares', 'spi_acs']
```
burnout87, line 90: This test, now passes

volodymyrss (approved):

tests/test_live.py
burnout87: any ideas @volodymyrss  ?

volodymyrss: what is the problem precisely?

burnout87: the list of instruments returned does not includ `magic`

volodymyrss: it should not. we do not provide magic.

PR #133, opened 2022-03-28
Get product type parent term

burnout87:


oda_api/api.py
```diff
@@ -60,13 +58,12 @@
...
 advice_logger = logging.getLogger("oda_api.advice")
 
-
 __all__ = ['Request', 'NoTraceBackWithLineNumber',
            'NoTraceBackWithLineNumber', 'RemoteException', 'DispatcherAPI']
 
 
 class Request(object):
-    def __init__(self,):
+    def __init__(self, ):
```
volodymyrss, line 66: why ","?

volodymyrss:
Uh, could we not mix substantial style changes in the same PR as features? It's hard to see scroll through all and spot the actual change.
See here https://github.com/oda-hub/doc-handbook/blob/master/style-guide.md
It'd be best best to freeze the style with CI check, and this https://github.com/oda-hub/doc-handbook/issues/5. But, later.

oda_api/api.py
burnout87: I honestly don't know, this must be some results of a merging, or an automatic change applied by my IDE (and I missed it)

burnout87:
> Uh, could we not mix substantial style changes in the same PR as features? 

I totally agree and have always followed the guideline, but this time was unintentional and the results of a merging or some automatic style adjuster from my IDE that I missed and I committed

volodymyrss:
> > Uh, could we not mix substantial style changes in the same PR as features?
> 
> I totally agree and have always followed the guideline, but this time was unintentional and the results of a merging or some automatic style adjuster from my IDE that I missed and I committed

Alright, as long as we agree on the intention, it's cool.
The IDE's do that sometimes, and even some pre-commit hooks do that. These behaviors should be opposed! But sometimes they can slip through...

volodymyrss (approved):
ok, I do not have a strong opinion about the content. Thanks

PR #138, opened 2022-05-03
job_id DataCollection attribute

burnout87:


oda_api/api.py
```diff
@@ -1275,6 +1276,14 @@ def __init__(self, data_list, add_meta_to_name=['src_name', 'product'], instrume
             self._p_list.append(data)
             self._n_list.append(var_name)
 
+    @property
+    def job_id(self):
```
volodymyrss, line 1280: we discussed it's useful to call it something else. e.g. `producing_job_id`.

I am not entire sure why setters/getters are useful here. Maybe they are?

burnout87: > we discussed it's useful to call it something else. e.g. `producing_job_id`.
> 

True! I had forgotten about it, thanks

> I am not entire sure why setters/getters are useful here. Maybe they are?

probably not really, I just applied the same approach used within the DispatcherAPI class, but yeah they're not really useful

volodymyrss: > > we discussed it's useful to call it something else. e.g. `producing_job_id`.
> 
> True! I had forgotten about it, thanks
> 
> > I am not entire sure why setters/getters are useful here. Maybe they are?
> 
> probably not really, I just applied the same approach used within the DispatcherAPI class, but yeah they're not really useful

I can be useful when getter should return something definitive before the value is set, especially when calculating the value. Here not so much.

volodymyrss (approved):

PR #139, opened 2022-05-03
update_data_product par for post_data_product_to_gallery function

burnout87:


volodymyrss (approved):

volodymyrss (dismissed):

burnout87:
Converting back to draft until https://github.com/oda-hub/dispatcher-app/issues/406 is resolved

burnout87:
Will also include the adaption with the changes introduced in https://github.com/oda-hub/dispatcher-app/pull/412

ferrigno:
I have updated some plot tools to create HTML rendering (bokeh for light curve and spectra, js9 for images).
There are commits at the end of July that are necessary to run the notebooks for the Gallery

burnout87:
> I have updated some plot tools to create HTML rendering (bokeh for light curve and spectra, js9 for images). There are commits at the end of July that are necessary to run the notebooks for the Gallery

Are the changes introduces with these commits used for the sole purpose of the gallery?

Because if not, then we should create a dedicated PR for this

ferrigno:
My changes are only for the gallery.

.gitignore
```diff
@@ -24,7 +24,12 @@ lib64
 CL.txt
 /venv
 #/doc/source/user_guide*
-#!/doc/source/user_guide/*rst
+/doc/source/user_guide/*.png
+/doc/source/user_guide/*.fits
+/doc/source/user_guide/*.root
+/doc/source/user_guide/ScienceWindowList-Copy1.ipynb
```
volodymyrss, line 30: why these lines?

volodymyrss:
that's a lot of changes to an important library. It will take a while too review. It's would be better to make incremental changes

burnout87:
> that's a lot of changes to an important library. It will take a while too review. It's would be better to make incremental changes

indeed, but it has to be merged, it has been reviewd in the past, and recently very few changfes have been introduced, and all for the gallery

.gitignore
burnout87: files generated with the execution of the example notebooks

volodymyrss: `ScienceWindowList-Copy1.ipynb` is generated? could in principle be, but where?

volodymyrss:
> > that's a lot of changes to an important library. It will take a while too review. It's would be better to make incremental changes
> 
> indeed, but it has to be merged, it has been reviewd in the past, and recently very few changfes have been introduced, and all for the gallery

where was it reviewed? few changes with respect to what? not with respect to the master, right?

this package is installed on every api client, it has to be done carefully. There are plenty of changes to `oda_api/api.py`. As gallery publishing is not available to most users, can could pay a bit less attention to it, but it should be then moved to a different file.

burnout87:
> > > that's a lot of changes to an important library. It will take a while too review. It's would be better to make incremental changes
> > 
> > 
> > indeed, but it has to be merged, it has been reviewd in the past, and recently very few changfes have been introduced, and all for the gallery
> 
> where was it reviewed? few changes with respect to what? not with respect to the master, right?
> 

respect to when the last review was done

> this package is installed on every api client, it has to be done carefully. There are plenty of changes to `oda_api/api.py`. As gallery publishing is not available to most users, can could pay a bit less attention to it, but it should be then moved to a different file.

this makes sense

ferrigno:
I have checkced the plot_tools.py and it looks OK.

burnout87:
> 
> could you link where the review was done? for me, the review is only done when the branch is merged on master. All else is in progress.
> 

ok it was approved some time ago, but quite some time has passed

> 
> if it's separated, we can see to make sure it does not introduce issues and technical debt in the core api

the more I look at the code and the more I realize that it is not so automatic...parts of the main class of `api.py` are used, for example, for making the requests to the dispatcher

burnout87:
> the more I look at the code and the more I realize that it is not so automatic...parts of the main class of `api.py` are used, for example, for making the requests to the dispatcher

and it was the approach we took since I started implementing this part

volodymyrss:
> > could you link where the review was done? for me, the review is only done when the branch is merged on master. All else is in progress.
> 
> ok it was approved some time ago, but quite some time has passed
> 
> > if it's separated, we can see to make sure it does not introduce issues and technical debt in the core api
> 
> the more I look at the code and the more I realize that it is not so automatic...parts of the main class of `api.py` are used, for example, for making the requests to the dispatcher

and this creates potential for issues, 
You can import and inherit.

volodymyrss:
> > the more I look at the code and the more I realize that it is not so automatic...parts of the main class of `api.py` are used, for example, for making the requests to the dispatcher
> 
> and it was the approach we took since I started implementing this part

way too much grew in this PR without merging. This is a critical component and before merging it should be done better.

volodymyrss:
> > > the more I look at the code and the more I realize that it is not so automatic...parts of the main class of `api.py` are used, for example, for making the requests to the dispatcher
> > 
> > 
> > and it was the approach we took since I started implementing this part
> 
> way too much grew in this PR without merging. This is a critical component and before merging it should be done better.

ok, I understand you are not here next week, so taking up a non-trivial transformation now is unfortunate. Do we really need to merge it now?..

burnout87:
> 
> ok, I understand you are not here next week, so taking up a non-trivial transformation now is unfortunate. Do we really need to merge it now?..

no, the cronjobs that are running on least use this branch, so no need to do it right now, but of course when I am back

burnout87:
> 
> and this creates potential for issues, You can import and inherit.

`api.py` no longer contains anything gallery-related, but everything is now moved to `gallery_api.py` where a class extending `DispatcherAPI` is now defined. Tests seem to be running fine

burnout87:
I took the liberty to address also #179

.gitignore
burnout87: this file was simply a test file, and ended up here as a mistake (most likely automatically added by the IDE)

oda_api/api.py
```diff
@@ -384,6 +381,17 @@ def generate_session_id(self, size=16):
         chars = string.ascii_uppercase + string.digits
         return ''.join(random.choice(chars) for _ in range(size))
 
+    @classmethod
```
volodymyrss, line 384: so we already a hash in `unique_response_json_fn` merged, they need to bereconciled somehow.

burnout87: I guess `unique_response_json_fn` can simply use the function `make_hash`

volodymyrss:
> > and this creates potential for issues, You can import and inherit.
> 
> `api.py` no longer contains anything gallery-related, but everything is now moved to `gallery_api.py` where a class extending `DispatcherAPI` is now defined. Tests seem to be running fine

This is good, thanks!

volodymyrss (approved):

PR #157, opened 2022-09-01
Missing module

volodymyrss:


oda_api/cli.py
```diff
@@ -72,6 +73,16 @@ def get(obj, instrument, product, argument, _discover_token):
                 for du in p.data_unit:
                     logger.info(">> %s", du.data)
 
+            for k, v in plot_tools.__dict__.items():
+                if not k.startswith('Oda'): continue
```
burnout87, line 77: why not just
```suggestion
                if k.startswith('Oda'):
```

volodymyrss: well it gets kind of deep. Though I had more stuff before, it's shorter now. Ok I changed

burnout87 (approved):

PR #163, opened 2022-10-12
functoinality to rewrite a discovered token

burnout87:


volodymyrss:
Fails for now!

burnout87:
Any ideas on why this is happening?
```python
RequestNotUnderstood: [ RequestNotUnderstood: RequestNotUnderstood():[ InstrumentQueryBackEnd : isgri ] constructor failed: ('!Error error ', OSError(28, 'No space left on device'), 'in creating dir', 'scratch_sid_IS0PEXDFEUDXX806_jid_2c8b65d7f62a0d7c_aliased') ]
```

when running the `test_plot_tools_notebook` test

burnout87:
Had to apply some changes regarding the default values for optional arguments as described  [here](https://github.com/hauntsaninja/no_implicit_optional)

burnout87:
> Any ideas on why this is happening?
> 
> ```python
> RequestNotUnderstood: [ RequestNotUnderstood: RequestNotUnderstood():[ InstrumentQueryBackEnd : isgri ] 
> constructor failed: ('!Error error ', OSError(28, 'No space left on device'), 'in creating dir', 
> 'scratch_sid_IS0PEXDFEUDXX806_jid_2c8b65d7f62a0d7c_aliased') ]
> ```
> 
> when running the `test_plot_tools_notebook` test

Issue seems to have been resolved

burnout87:
Can you please review again @volodymyrss ? So I can merge

volodymyrss:
Please wait until review.

oda_api/token.py
```diff
@@ -66,6 +65,34 @@ def decode_oauth2_token(token: str):
...
+    ]:
+        if n in token_discovery_methods:
+            current_token = discover_token(token_discovery_methods=n)
+            if current_token is not None and current_token != '':
+                if n == "environment variable ODA_TOKEN":
+                    environ['ODA_TOKEN'] = new_token
+                    break
+                elif n == "file in current directory":
+                    with open(path.join(getcwd(), ".oda-token"), 'w') as ft:
+                        ft.write(new_token)
```
volodymyrss, line 88: this can potentially overwrite a useful token.

oda_api/token.py
```diff
@@ -66,6 +65,34 @@ def decode_oauth2_token(token: str):
     # usually comes in cookies['_oauth2_proxy']
     return json.loads(base64.b64decode(token.split(".")[0]+"=").decode())
 
+
+def rewrite_token(new_token,
+                  token_discovery_methods=(
```
volodymyrss, line 70: I am not sure discovery is applicable when the token is being written.
When it's being found, it find the first usable one.

When writing, there are other options.
Among them, I think it's good to explicitly indicate where to store the token, not just bool.

Also, when token is being overwritten, report what is the difference between old and new token. Maybe even save the old token.

oda_api/token.py
```diff
@@ -66,6 +65,34 @@ def decode_oauth2_token(token: str):
...
+                          "environment variable ODA_TOKEN",
+                          "file in current directory",
+                          "file in home")):
+
+    for n in [
+        "environment variable ODA_TOKEN",
```
volodymyrss, line 76: since these modes are now in two places, and even maybe will be exposed to the user, they could be replaced with Enum.

volodymyrss (changes requested):
few points about overwritting

oda_api/token.py
burnout87: > I am not sure discovery is applicable when the token is being written. When it's being found, it find the first usable one.
> 
> When writing, there are other options. Among them, I think it's good to explicitly indicate where to store the token, not just bool.
>

You mean having the user explicitly specify where the new token will be written?

volodymyrss: Yes. `write_token=False`, `write_token=token.HOME_FILE`, etc.

Also, when overwriting, to check if the new one is actually better. I.e. has at least the same roles and at least the same remaining lifetime.

burnout87: The refresh_token method allows now to specify where to store the new token.

Also, the old token is stored and afterwards the previous "location" used is emptied, in order to guarantee that the new, refreshed, token, get found properly.

burnout87:
@dsavchenko the `test_bin_image` is failing, did you introduce it? Any ideas of the cause?

dsavchenko:
Strange. It's unstable, passed after restart. 
Seems that the bytes get corrupted, but only sometimes 
https://github.com/oda-hub/oda_api/issues/169
I will try to investigate.

burnout87:
> Strange. It's unstable, test passed after restart. Seems that the bytes get corrupted, but only sometimes #169 I will try to investigate.

thanks!

burnout87:
> Strange. It's unstable, test passed after restart. Seems that the bytes get corrupted, but only sometimes #169 I will try to investigate.

Indeed they all passed now...

dsavchenko:
Seems that https://github.com/oda-hub/oda_api/pull/170 is a fix

burnout87:
> Seems that #170 is a fix

Indeed it does

oda_api/api.py
```diff
@@ -331,6 +331,32 @@ def inspect_state(self, job_id=None):
...
+
+            r = requests.get(os.path.join(self.url, 'refresh_token'),
+                             params=params)
+
+            if r.status_code == 200:
+                refreshed_token = r.text
+                if write_token:
+                    token_write_method_enum = None
+                    if token_write_method is not None:
+                        token_write_method_enum = oda_api.token.TokenAccessMethods[str.upper(token_write_method)]
```
volodymyrss, line 351: this is not needed, users can pass enum

oda_api/token.py
```diff
@@ -62,57 +69,125 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+                  ):
+    current_token, discover_method = discover_token_and_method(allow_invalid=True)
+    current_decoded_token = decode_oda_token(current_token, allow_invalid=True)
+    new_decoded_token = decode_oda_token(new_token, allow_invalid=True)
+
+    current_decoded_token_expires_in_s = current_decoded_token['exp'] - time.time()
+    new_decoded_token_expires_in_s = new_decoded_token['exp'] - time.time()
+
+    if new_decoded_token_expires_in_s < current_decoded_token_expires_in_s:
+        logger.info("the new token has will expire before the current one")
```
volodymyrss, line 101: if this happens, something is probably really wrong. I'd really return right here and not write, unless "force=True" is passed.

oda_api/token.py
```diff
@@ -62,57 +69,125 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+
+    current_decoded_token_roles = get_token_roles(current_decoded_token)
+    new_decoded_token_roles = get_token_roles(new_decoded_token)
+    if set(current_decoded_token_roles) - set(new_decoded_token_roles) != set():
+        logger.info("The new token has a different set of roles than the current one:\n"
```
volodymyrss, line 106: I'd say if new roles include all old roles - logger.warning("new token has more roles than the existing one: xxxx"). If new roles are fewer - a warning and return, unless "force=True".

oda_api/token.py
```diff
@@ -62,57 +69,125 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+                    f"roles current token: {current_decoded_token_roles}\n"
+                    f"roles new token: {new_decoded_token_roles}")
+
+    if token_write_method is not None:
+        # TODO perhaps use a more appropriate location
+        with open("old-oda-token_" + str(time.time()), 'w') as ft:
+            ft.write(current_token)
+        if discover_method is not None:
+            if discover_method == TokenLocation.ODA_ENV_VAR:
```
volodymyrss, line 115: this exterminates all tokens and can be quite dangerous. What if some of them were better. I have it happen sometimes: I may be using underprivileged user, but I have high-priveged token stored.

I suggest to not silently overwrite. See next comment.

oda_api/token.py
```diff
@@ -62,57 +69,125 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+                environ.pop('ODA_TOKEN', None)
+            elif discover_method == TokenLocation.FILE_CUR_DIR:
+                if os.path.exists(path.join(getcwd(), ".oda-token")):
+                    os.remove(path.join(getcwd(), ".oda-token"))
+            elif discover_method == TokenLocation.FILE_HOME:
+                if os.path.exists(path.join(environ["HOME"], ".oda-token")):
+                    os.remove(path.join(environ["HOME"], ".oda-token"))
+
+        if token_write_method == TokenLocation.ODA_ENV_VAR:
```
volodymyrss, line 124: actually, could be good to have `token_write_methods`, i.e. a tuple. 

To avoid the problem that sometimes the user writes something which will not be used in the discovery, do just that: check that the token is dicovered after it's written.

volodymyrss:
many comments, but this is a bit tricky and I stumbled on some details with these sort of things in oda-api and elsewhere. It's crutial for good user experience that we think it through taking into account our practical experience.

oda_api/token.py
volodymyrss: if it's not discovered: revert.

oda_api/token.py
```diff
@@ -62,57 +69,125 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+            environ['ODA_TOKEN'] = new_token
+        elif token_write_method == TokenLocation.FILE_CUR_DIR:
+            with open(path.join(getcwd(), ".oda-token"), 'w') as ft:
+                ft.write(new_token)
```
volodymyrss, line 128: whenever you write in these case, you need to be sure the permissions are `400`.

oda_api/token.py
burnout87: > actually, could be good to have `token_write_methods`, i.e. a tuple.
> 
What does it mean?

volodymyrss: User may want to write to multiple locations at once. I.e. it could be 
`token_write_methods: Union[Tuple[TokenLocation], TokenLocation]`

oda_api/token.py
burnout87: > this exterminates all tokens and can be quite dangerous. What if some of them were better. I have it happen sometimes: I may be using underprivileged user, but I have high-priveged token stored.
> 

In such a situation (old token better than the new one) there is the check on the roles, and that partally verifies exactly this

oda_api/token.py
burnout87: > if it's not discovered: revert.

You mean remove the newly written token? And so keep using the current/old one?

oda_api/api.py
```diff
@@ -331,6 +332,29 @@ def inspect_state(self, job_id=None):
         else:
             raise RuntimeError(r.text)
 
+    def refresh_token(self,
+                      write_token=False,
+                      # TODO does this make sense as a default one?
```
volodymyrss, line 337: maybe, let's keep this

oda_api/api.py
```diff
@@ -331,6 +332,29 @@ def inspect_state(self, job_id=None):
...
+            if r.status_code == 200:
+                refreshed_token = r.text
+                if write_token:
+                    oda_api.token.rewrite_token(refreshed_token, token_write_methods=token_write_methods)
+
+                return refreshed_token
+            else:
+                raise RuntimeError(r.text)
+        else:
+            raise RuntimeError("failed to discover token with any known method")
```
volodymyrss, line 356: + ", unable to refresh"

oda_api/token.py
```diff
@@ -62,57 +68,154 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
 
-#TODO: move to dynaconf
-def discover_token(
+
+def get_token_roles(decoded_token):
+    # extract role(s)
+    roles = None
+    if isinstance(decoded_token['roles'], str):
+        roles = decoded_token['roles'].split(',') if 'roles' in decoded_token else []
+        roles[:] = [r.strip() for r in roles]
```
volodymyrss, line 82: why `roles[:]`?

oda_api/token.py
```diff
@@ -62,57 +68,154 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
     return json.loads(base64.b64decode(token.split(".")[0]+"=").decode())
 
-#TODO: move to dynaconf
-def discover_token(
+
+def get_token_roles(decoded_token):
+    # extract role(s)
+    roles = None
+    if isinstance(decoded_token['roles'], str):
+        roles = decoded_token['roles'].split(',') if 'roles' in decoded_token else []
```
volodymyrss, line 81: we just check that `decoded_token['roles']` is `str`, it's definitily in there :)

oda_api/token.py
```diff
@@ -62,57 +68,154 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+        new_decoded_token_expires_in_s = new_decoded_token['exp'] - time.time()
+
+        if new_decoded_token_expires_in_s < current_decoded_token_expires_in_s:
+            warning_msg = "The new token will expire before the current one"
+            if force_rewrite:
+                warning_msg += ", but it will be used"
+                logger.warning(warning_msg)
+            else:
+                raise RuntimeError("Expiration time of the refreshed token is lower than "
+                                   "the currently available one")
```
volodymyrss, line 108: ", please pass force=True to overwrite"

oda_api/token.py
volodymyrss: Right, good point, we need to check that all other token settings are the same. Refresh should only change the duration - that's its meaning. Maybe roles in a rare case.

oda_api/token.py
```diff
@@ -62,57 +68,154 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+                ft.write(current_token)
+
+        if isinstance(token_write_methods, TokenLocation):
+            token_write_methods = token_write_methods,
+
+        if discover_method is not None:
+            if discover_method == TokenLocation.ODA_ENV_VAR:
+                environ.pop('ODA_TOKEN', None)
+            elif discover_method == TokenLocation.FILE_CUR_DIR:
+                if path.exists(path.join(getcwd(), ".oda-token")):
```
volodymyrss, line 141: how can it not exist when it was discovered there?
If it does not exist - we should raise RuntimeError because it means there is some internal bug

oda_api/token.py
volodymyrss: Anyway, I still think this is way too dangerous and in case of any bug causes. Maybe this can be enabled with an extra option `discard_all_existing_tokens` which should be off by default.

oda_api/token.py
```diff
@@ -62,57 +68,154 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+                environ['ODA_TOKEN'] = new_token
+            elif token_write_method == TokenLocation.FILE_CUR_DIR:
+                with open(path.join(getcwd(), ".oda-token"), 'w') as ft:
+                    ft.write(new_token)
+                chmod(path.join(getcwd(), ".oda-token"), 0o400)
+            elif token_write_method == TokenLocation.FILE_HOME:
+                with open(path.join(environ["HOME"], ".oda-token"), 'w') as ft:
+                    ft.write(new_token)
+                chmod(path.join(environ["HOME"], ".oda-token"), 0o400)
+
```
volodymyrss, line 158: here we can add a sanity check that `discover_token` returns the freshly written one. For consistency

volodymyrss:
By the way, we also really need to have a help for this in readthedocs. Maybe in existing notebook  there https://oda-api.readthedocs.io/en/latest/user_guide/Authentication.html#Storing-your-token-locally

oda_api/token.py
burnout87: By reverting, so keep using the old one, don't we risk to have the newly refreshed token ending up not being used?

oda_api/token.py
burnout87: the file has to be set to 600 before any write, otherwise it won't be impossible to write anything (if the file already exists of course)

oda_api/token.py
burnout87: Sanity check added, and a warning is raised in case the discovered one is different from the refreshed one

oda_api/token.py
burnout87: > I am not sure discovery is applicable when the token is being written. When it's being found, it find the first usable one.
> 

The token to refresh can now be passed as argument, if that's not the case, it will be searched and discovered. Also to me it makes more sense.

> 
> Also, when token is being overwritten, report what is the difference between old and new token. Maybe even save the old token.

A dedicated comparison function now actually checks the fields of the payloads of the old token and the refreshed one, differences are reported and, if necessary, the force the writing.

oda_api/token.py
burnout87: `discard_all_existing_tokens` is now an argument of the function, False by default.

oda_api/token.py
burnout87: discussing the extensively down in other comments

oda_api/token.py
```diff
@@ -62,57 +71,302 @@ def decode_oda_token(token: str, secret_key=None, allow_invalid=False) -> dict:
...
+                if path.exists(path.join(getcwd(), ".oda-token")):
+                    chmod(path.join(getcwd(), ".oda-token"), 0o600)
+                with open(path.join(getcwd(), ".oda-token"), 'w') as ft:
+                    ft.write(new_token)
+                chmod(path.join(getcwd(), ".oda-token"), 0o400)
+            elif token_write_method == TokenLocation.FILE_HOME:
+                if path.exists(path.join(environ["HOME"], ".oda-token")):
+                    chmod(path.join(environ["HOME"], ".oda-token"), 0o600)
+                with open(path.join(environ["HOME"], ".oda-token"), 'w') as ft:
+                    ft.write(new_token)
```
volodymyrss, line 302: could you do a `logger.info` every time you write? Very useful to see.

volodymyrss:
Not all tokens can be refreshed, you need a role. Could you please add it in the doc?

burnout87:
> Not all tokens can be refreshed, you need a role. Could you please add it in the doc?

in the readthedocs, as well as [here](https://github.com/oda-hub/doc-multi-user/blob/main/plan-roles-users.md#actively-used-common-roles) ?

volodymyrss:
thanks! could we maybe move the paragraph in the beginning of the section? For most users there is no point of reading it and then realizing they can not do it.

Also, could we group sections a bit? By setting lower level titles. E.g. all token-related sections can be a subsection of "using personal access tokens". I am getting a bit lost in so many unstructured sections.

image

burnout87:
> thanks! could we maybe move the paragraph in the beginning of the section? For most users there is no point of reading it and then realizing they can not do it.
> 
> Also, could we group sections a bit? By setting lower level titles. E.g. all token-related sections can be a subsection of "using personal access tokens". I am getting a bit lost in so many unstructured sections.
> 

Good point, I re-organized a bit the structure of the various sections.

volodymyrss (approved):

PR #164, opened 2022-10-25
fix lc time format

dsavchenko:


oda_api/data_products.py
```diff
@@ -723,7 +723,7 @@ def from_arrays(cls,
         # TODO: possibility for other time units
         # TODO: add time-related keywords to header (OGIP)
         units_dict = {'TIME': 'd'}
-        if [isinstance(x, astropy.time.Time) for x in times]:
+        if any(isinstance(x, astropy.time.Time) for x in times):
```
burnout87, line 726: Just for my understanding, this was failing in the case, for example:
```python
times=[59630.3, 59630.5, 59630.7]
```

burnout87 (approved):

oda_api/data_products.py
dsavchenko: Exactly. Because bool([False, False, False]) is True, but time_format isn't being set inside this block, so it was failing.

Issue #169, opened 2022-11-23
PictureProduct: encoding-decoding unstable, fails sometimes


dsavchenko:
Could be also the case for BinaryData

PR #170, opened 2022-11-23
Fix picture encode-decode

dsavchenko:
Fix for #169


burnout87 (approved):
Issue appears only for `PictureProduct` objs?

PR #172, opened 2022-12-05
fix image wcs header

dsavchenko:


oda_api/data_products.py
```diff
@@ -763,7 +763,7 @@ def from_arrays(cls,
                                  meta_data = meta_data,
                                  data_header = data_header,
                                  hdu_type = 'bintable',
-                                 name = 'LIGHTCURVE')],
+                                 name = 'LC')],
```
dsavchenko, line 766: This is unrelated. Anyway, the name is arbitrary

volodymyrss (approved):

PR #174, opened 2022-12-08
show status comments

dsavchenko:


volodymyrss (changes requested):
ok! but please test

dsavchenko:
> ok! but please test

Unrelated, but live test is failing now for isgri with `No space left on device`!

volodymyrss:
> > ok! but please test
> 
> Unrelated, but live test is failing now for isgri with `No space left on device`!

https://github.com/oda-hub/dispatcher-app/issues/438

dsavchenko:
'warning' field is not used anywhere, so I don't test it

volodymyrss:
> 'warning' field is not used anywhere, so I don't test it

It's not used in any of the plugins? It was. 
Front-end supports it, and you add support here, so we need a test or discard it completely, else it can cause confusion.

And we need it in some cases. E.g. I think we had it when the result was empty.

dsavchenko:
It's defined in schema 
https://github.com/oda-hub/dispatcher-app/blob/fc3203586097a9840c2a4ab3a8694f99e2b160d2/cdci_data_analysis/flask_app/schemas.py#L22 
but its usage is undocumented, and I think we really didn't use it. 
I also don't see its support in frontend.

volodymyrss:
> It's defined in schema https://github.com/oda-hub/dispatcher-app/blob/fc3203586097a9840c2a4ab3a8694f99e2b160d2/cdci_data_analysis/flask_app/schemas.py#L22 but its usage is undocumented

Schema is sufficient description to consider it set
Undocumented comparing to what, other dispatcher classes? None of these development interfaces are documented.

>, and I think we really didn't use it. I also don't see its support in frontend.

It was used, and if it is not now, it needs to be restored.
If necessary, we can discuss it on Monday.

dsavchenko:
> Undocumented comparing to what

I mean, description is empty. For other fields, it describes intended usage.

volodymyrss:
> > Undocumented comparing to what
> 
> I mean, description is empty. For other fields, it describes intended usage.

Ah ok. It was meant to be, and I think it was, shown in a more threatening way, as opposed to comment.

We can also deal with it elsewhere https://github.com/oda-hub/dispatcher-app/issues/484

Here, adding the code but not the test is inconsistent.
Could you at least add a TODO, then we can address it later as needed.

volodymyrss (approved):

dsavchenko:
Temporarily disabled it. Also edited old-style message for missing args (let's keep it for now)

PR #175, opened 2022-12-12
plot image using astropy wcs

ferrigno:
It uses astropy for image projection


volodymyrss (approved):
This looks so much simpler, if it works it's good

PR #180, opened 2023-01-27
Plottools no sources exception

volodymyrss:
else it crashes on me


ferrigno:
I think that this is trying to merge an old version of plot tools, I removed this exception with some if in my current master

ferrigno:
I should have uploaded to this branch the current version of plot_tools.py

volodymyrss:
with this updated version, we do not need this PR, but still would be useful to figure out what happenned.

PR #182, opened 2023-01-31
local cache for the results

volodymyrss:


oda_api/api.py
```diff
@@ -1317,6 +1335,33 @@ def set_api_code(query_dict, url="www.astro.unige.ch/mmoda/dispatch-data"):
 
         return _cmd_
 
+
+    def save_result(self, response_json):
+        fn = self.unique_response_json_fn
+
+        os.makedirs(os.path.dirname(fn), exist_ok=True)
+
+        json.dump(response_json, gzip.open(fn, "wt"))
```
dsavchenko, line 1344: Generally, I always prefer using `with` when opening file

oda_api/api.py
```diff
@@ -1317,6 +1335,33 @@ def set_api_code(query_dict, url="www.astro.unige.ch/mmoda/dispatch-data"):
...
+        logger.info('saved result in %s', fn)
+
+
+    def load_result(self):
+        fn = self.unique_response_json_fn
+        logger.info('trying to load result from %s', fn)
+
+        t0 = time.time()
+        r = json.load(gzip.open(fn, 'rb'))
```
dsavchenko, line 1353: I see it works (passes test) but it's not clear to me, why writing with `text` mode and reading with `bytes`

oda_api/api.py
volodymyrss: yeah, well, it adds an extra line and an extra variable. Ultimately the file would get closed.
But ok, it's safer to make sure it's closed here.

oda_api/api.py
volodymyrss: it's just how `json` module works. Both text also works.

PR #194, opened 2023-03-22
List products per source

burnout87:


volodymyrss:
> > Ok. I think it's best if you mention me in the PR if needed, this sends me an email. This is how it's done where I saw.
> 
> Ok, I'll keep that in mind. BTW you don't get emails in case review is requested? I usually have them

I do, but I can not always reply. Please remind if needed, else I only come back to it once I find a moment in an ordinary schedule.

volodymyrss (approved):

PR #199, opened 2023-06-20
passing token to the call of some methods

burnout87:

PR #200, opened 2023-06-27
Keep names of all products + new BinaryProduct

dsavchenko:


burnout87:
Live test is failing in the same way as https://github.com/oda-hub/oda_api/pull/199#issuecomment-1604285566

oda_api/api.py
```diff
@@ -1250,28 +1256,38 @@ def from_response_json(cls, res_json, instrument, product):
...
 
         if 'binary_data_product_list' in res_json['products'].keys():
-            data.extend([BinaryData().decode(d)
-                         for d in res_json['products']['binary_data_product_list']])
+            try:
+                data.extend([BinaryProduct.decode(d)
+                             for d in res_json['products']['binary_data_product_list']])
+            except:
+                data.extend([BinaryData().decode(d)
+                             for d in res_json['products']['binary_data_product_list']])
```
burnout87, line 1264: why is this needed?

dsavchenko: The except case is just for backward compatibility, New BinaryProduct will be used in general

oda_api/data_products.py
```diff
@@ -240,9 +276,9 @@ def _check_dict(self,_kw):
 
 
     @classmethod
-    def from_fits_hdu(cls,hdu,name=''):
+    def from_fits_hdu(cls,hdu,name=None):
 
-        if name=='':
+        if not name:
```
burnout87, line 281: I think PEP8 recommends to always use `is` operator for checking against `None`

burnout87 (commented):
few  comments/questions

oda_api/data_products.py
dsavchenko: Here I wanted to account for both None and empty string (compatibility again, but not sure if it's really needed)

burnout87: then  I'd put:

```python
if name is not None or name == '':
```

burnout87 (dismissed):

oda_api/api.py
```diff
@@ -1250,28 +1257,36 @@ def from_response_json(cls, res_json, instrument, product):
                          for d in res_json['products']['numpy_data_product_list']])
 
         if 'binary_data_product_list' in res_json['products'].keys():
-            data.extend([BinaryData().decode(d)
-                         for d in res_json['products']['binary_data_product_list']])
+            try:
+                data.extend([BinaryProduct.decode(d)
+                             for d in res_json['products']['binary_data_product_list']])
+            except:
```
volodymyrss, line 1263: could you catch something more specific?

volodymyrss:
what's with Untitled.ipynb

dsavchenko:
> what's with Untitled.ipynb

It was added by mistake

volodymyrss (approved):

PR #203, opened 2023-07-11
Log to fits hdu function

burnout87:


burnout87:
How do you see having a setter and a getter property for the `name` attribute? @dsavchenko 

It is now applied to the class `NumpyDataUnit`, but can be easily extended to the other classes

dsavchenko:
The idea of #200 was that the product should have no name unless explicitly set (otherwise, either this default name was substituting the name of the variable or should have been explicitly reset in the plugin)

Therefore, no need to change it in other products. The NumpyDataUnit is special case, it represents fits extension, which should always have name (and it is not used as a product by itself).

burnout87:
> The idea of #200 was that the product should have no name unless explicitly set (otherwise, either this default name was substituting the name of the variable or should have been explicitly reset in the plugin)
> 
> Therefore, no need to change it in other products. The NumpyDataUnit is special case, it represents fits extension, which should always have name (and it is not used as a product by itself).

Ok then, the changes I included assign `table` as name in the case this is None, ONLY for `NumpyDataUnit` objects.

I also extended an extended test that verifies that this is set

volodymyrss (approved):

dsavchenko (approved):

PR #207, opened 2023-07-28
fix the case for booleans in the `api_code`

burnout87:


volodymyrss (approved):
ok, although I think there is a better way to produce python dict literal string directly, since we have json so let's keep it.

PR #208, opened 2023-09-05
json from stdlib instead of flask

dsavchenko:
following https://gitlab.renkulab.io/astronomy/mmoda/photoz-euclid/-/issues/7


volodymyrss (approved):

dsavchenko:
@volodymyrss there is some problem building the docs. I never touched it, could you have a look?

PR #209, opened 2023-09-05
almost empty rtd config

volodymyrss:


volodymyrss:
FYI @burnout87 also, this might be relevant elsewhere

PR #225, opened 2023-11-07
217 add functionality to send callback messages

okolo:
Progress reporter API was added


oda_api/api.py
```diff
@@ -1339,3 +1340,48 @@ def from_response_json(cls, res_json, instrument, product):
...
+        return self._callback is not None
+
+    def report_progress(self, stage: str=None, progress: int=50, substage: str=None, subprogress: int=None, message:str=None):
+        """
+        Report progress via callback URL
+        :param stage: current stage description string
+        :param stage: current stage progress in %
+        :param stage: current substage description string
+        :param stage: current substage progress in %
+        :param message: message to pass
```
burnout87, line 1367: I guess you missed the arg names

oda_api/api.py
```diff
@@ -1339,3 +1340,48 @@ def from_response_json(cls, res_json, instrument, product):
...
+        """
+        callback_payload = {k: str(v) for k, v in locals().items() if v is not None and k != 'self'}
```
burnout87, line 1369: this is going to put the entire content of locals as a payload for the request, is it really needed? can it be filtered to remove not-needed items?

burnout87:
Is it also possible to add a test?

oda_api/api.py
dsavchenko: I think it's ok, as locals() in this content are only method parameters

dsavchenko: I think it's ok, as locals() in this content are only method parameters

oda_api/api.py
```diff
@@ -1339,3 +1340,48 @@ def from_response_json(cls, res_json, instrument, product):
...
 
         return d
+
+class ProgressReporter(object):
+    """
+    The class allows to report task progress to end user
+    """
+    def __init__(self):
+        self._callback = None
+        callback_file = ".oda_api_callback"  # perhaps it would be better to define this constant in a common lib
```
dsavchenko, line 1350: To be more general, it could be the init parameter with default value. Also, probably good to have another parameter to be able to set the url directly. (Even though the current usage is by passing the file of fixed name)

oda_api/api.py
```diff
@@ -1339,3 +1340,48 @@ def from_response_json(cls, res_json, instrument, product):
...
+
+        logger.info('will perform callback: %s', self._callback)
+
+        if re.match('^file://', self._callback):
+            with open(self._callback.replace('file://', ''), "w") as f:
+                json.dump(callback_payload, f)
+            logger.info('stored callback in a file %s', self._callback)
+
+        elif re.match('^https?://', self._callback):
+            r = requests.get(self._callback, params=callback_payload)
```
dsavchenko, line 1383: Need try-except here to account for the situation when url is unreachable

oda_api/api.py
```diff
@@ -1339,3 +1340,48 @@ def from_response_json(cls, res_json, instrument, product):
...
+        callback_payload = {k: str(v) for k, v in locals().items() if v is not None and k != 'self'}
+
+        if not self.enabled:
+            logger.info('no callback registered, skipping')
+            return
+
+        logger.info('will perform callback: %s', self._callback)
+
+        if re.match('^file://', self._callback):
+            with open(self._callback.replace('file://', ''), "w") as f:
```
dsavchenko, line 1378: Would be good to ensure that basedir is available and create the directory tree if not (even though "write callback to file" mode is only used in tests)

oda_api/api.py
```diff
@@ -1339,3 +1340,48 @@ def from_response_json(cls, res_json, instrument, product):
...
+    def report_progress(self, stage: str=None, progress: int=50, substage: str=None, subprogress: int=None, message:str=None):
+        """
+        Report progress via callback URL
+        :param stage: current stage description string
+        :param progress: current stage progress in %
+        :param substage: current substage description string
+        :param subprogress: current substage progress in %
+        :param message: message to pass
+        """
+        callback_payload = {k: str(v) for k, v in locals().items() if v is not None and k != 'self'}
```
dsavchenko, line 1369: @burnout87 @volodymyrss I'm not sure, is it needed for dispatcher that `"action": "progress"` is in the callback payload?

volodymyrss: Do you want to refer to the code where this is parsed?

burnout87: dispatcher uses it here: https://github.com/oda-hub/dispatcher-app/blob/5b8c558a8d107d94774813f00402ca92ebeca2d8/cdci_data_analysis/flask_app/dispatcher_query.py#L1069-L1074

dispatcher needs to know that, and will it for email/matrix messages

oda_api/api.py
volodymyrss: If you make a change, @okolo , please respond and resolve so we know, ok?

oda_api/api.py
volodymyrss: > this is going to put the entire content of locals as a payload for the request, is it really needed? can it be filtered to remove not-needed items?

I agree with @burnout87 that this kind of code-shortening trickery, although inventive, is prone to mistakes - what if someone adds any simple assignment, in some check?
You do spell out all of the parameters twice - in function spec and in the doc string. Adding one more time does not hurt here.

Or, if you want shorter, use kwargs?
Or, inspect function parameters? 

But really, it's not too much of a duplication to introduce error-prone complexity.
Even if I indeed unironically appreciate inventive code!

oda_api/api.py
volodymyrss: What do you mean @dsavchenko , do what with except?
If internal exception raises it can be shown as is, no need to re-raise.
Maybe a retry can be useful, but through requests, not custom.

oda_api/api.py
volodymyrss: Sorry, not sure about this at all. Not too good to create directories following a typo.

oda_api/api.py
volodymyrss: This is all somehow all outside intended usage. Interface with nb2workflow should be quite fixed. In can be changed in the future if needed, but there is no use case at the moment, is it, @dsavchenko ?

oda_api/api.py
volodymyrss: "progress" will be shown in frontend and API response. People do distinguish "progress" from "unknown", visually :) This is for people.

burnout87: it does not, but it will appear within the email/matrix message, but of course it will not disrupt the overall execution of the job

oda_api/api.py
dsavchenko: I agree that there is no use case at the moment. It was just a proposal

oda_api/api.py
dsavchenko: I mean this progress-reporting callback is not critical for the workflow itself, so why not log an error and allow workflow to continue?

oda_api/api.py
dsavchenko: Probably, you are right

oda_api/api.py
volodymyrss: This whole PR is to help user understand what's happening, we do not want to have user see numerous "unknown" word as sign of progress :)

oda_api/api.py
volodymyrss: If communication with the dispatcher is not possible,  "done" message will not arrive and the user experience is "hanging request" which is bad.
Silently failing user-requested code throughout the execution is also in general unfortunate.
Actually I think this should end up in a sentry message, can we do it?

oda_api/api.py
volodymyrss: We keep it in mind!

oda_api/api.py
dsavchenko: This is not about "done" status, but progress report. "done" is sent by nb2workflow upon completion. The lack of intermediate progress report is not so critical. 

Sentry is an option, but then we need to pass sentry credentials to the backend, which runs user-provided code, this is a security issue

volodymyrss: > This is not about "done" status, but progress report. "done" is sent by nb2workflow upon completion. The lack of intermediate progress report is not so critical.

The most realistic reason it would fail is due to problem accessing dispatcher. The same problem would prevent "done".
If there is some bug, or the callback reference file is overwritten, or something like that - the functionality intended by the user will working, and the right course of action is to let us and the user know, not silently ignore. If there is a warning inside, some users will see some dropped progress messages and it will be hard to know when this happens. That's not a good experience.

> Sentry is an option, but then we need to pass sentry credentials to the backend, which runs user-provided code, this is a security issue

It is really not https://docs.sentry.io/product/sentry-basics/concepts/dsn-explainer/#dsn-utilization
The problem can be with alert flooding, which may be challenging in some cases, but it is not the case for us at the moment.
Note that every web tracking framework, e.g. google analytics, has a similar issue.

oda_api/api.py
```diff
@@ -1339,3 +1340,49 @@ def from_response_json(cls, res_json, instrument, product):
...
+        Report progress via callback URL
+        :param stage: current stage description string
+        :param progress: current stage progress in %
+        :param substage: current substage description string
+        :param subprogress: current substage progress in %
+        :param message: message to pass
+        """
+        callback_payload = dict(stage=stage, progress=progress, substage=substage, subprogress=subprogress, message=message)
+        callback_payload = {k: v for k, v in callback_payload.items() if v is not None}
+
```
dsavchenko, line 1371:
```suggestion
        callback_payload['action'] = 'progress'
```

tests/test_progress_report.py
```diff
@@ -0,0 +1,34 @@
...
+
+callback_file = ".oda_api_callback"
+
+def test_progress_reporter_disabled():
+    if os.path.isfile(callback_file):
+        os.remove(callback_file)
+    # if callback is not available
+    pr = ProgressReporter()
+    assert not pr.enabled
+    pr.report_progress(stage='simulation', progress=50, substage='spectra', subprogress=30, message='some message')
```
burnout87, line 13: perhaps it should be verified that basically nothing happens?

burnout87 (changes requested):
Should an instance of the `ProgressReporter` be a member of any `DispatcherAPI` object?

I think it would be more clean and transparent

volodymyrss:
Please add docs too, @okolo .  See [where they appear for this PR](https://oda-api--225.org.readthedocs.build/en/225/) .

tests/test_progress_report.py
okolo: I am not sure what precisely you suggest to verify. In the test_progress_reporter_disabled() we check that the progress reporter doesn't fail if callback URL is not available. In this case the report_progress does nothing.

burnout87: what do you expect to see after having executed called`report_progress`? this should be investigated

volodymyrss:
> Should an instance of the `ProgressReporter` be a member of any `DispatcherAPI` object?
> 
> I think it would be more clean and transparent

Because it's a different way to access dispatcher, it may be a separate class.

oda_api/api.py
dsavchenko: We can probably keep it "resolved" here as the basic implementation, but I added an Issue https://github.com/oda-hub/oda_api/issues/229

burnout87 (approved):

dsavchenko (approved):

PR #226, opened 2023-11-09
token clean-up `test_instruments`

burnout87:


burnout87:
interestingly, I had already developed a fixture specific for cleaning up the tokens from the environment

volodymyrss (approved):

PR #237, opened 2024-01-29
allow digits and some characters in url path

dsavchenko:
These are definitely allowed by standards. Hope we won't encounter a need to use a path with url-encoded / some more allowed symbols

### DIFF
--- a/oda_api/misc_helpers.py
+++ b/oda_api/misc_helpers.py
@@ -7,7 +7,7 @@ regex_url = re.compile(
                 r'^https?://'  # http:// or https://
                 r'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'  # domain...
                 r'(?::\d+)?'  # optional port
-                r'(?:/[a-z0-9.~/-]*)?$', re.IGNORECASE)
+                r'(?:/[a-z0-9/-]*)?$', re.IGNORECASE)
 
 
 def validate_url(url):
--- a/oda_api/misc_helpers.py
+++ b/oda_api/misc_helpers.py
@@ -7,7 +7,7 @@ regex_url = re.compile(
                 r'^https?://'  # http:// or https://
                 r'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'  # domain...
                 r'(?::\d+)?'  # optional port
-                r'(?:/[a-z-/]*)?$', re.IGNORECASE)
+                r'(?:/[a-z0-9.~/-]*)?$', re.IGNORECASE)
 
 
 def validate_url(url):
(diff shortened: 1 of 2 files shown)


burnout87:
> These are definitely allowed by standards. Hope we won't encounter a need to use a path with url-encoded / some more allowed symbols

Would it be possible to provide an example/test ?

volodymyrss (changes requested):
This function is for validating dispatcher URLs specifically, which are narrower for a variety of reasons.
And it's helpful to notice likely incorrect URLs.

If you have new cases, please do modify, but also add some tests in there:

https://github.com/oda-hub/oda_api/blob/8316675687bb75627ac8a3b352312bc387091d3b/tests/test_basic.py#L249

dsavchenko:
I had created a somewhat obfuscated url to directly access the staging dispatcher.
It contains digits in the path part. I can change it, of course, but I see no reason for path to not contain digits. Tilde and dot are probably overshoot, I agree.

volodymyrss:
> I had created a somewhat obfuscated url to directly access the staging dispatcher. It contains digits in the path part. I can change it, of course, but I see no reason for path to not contain digits. Tilde and dot are probably overshoot, I agree.

Sure, add numbers.

volodymyrss (approved):

volodymyrss:
> The failed live-test looks completely unrelated. Is it some transient issue with isgri?

Probably. Strange. By the way, something very strange is going on with testing all UNIGE sites since 3 days, even very basic ones, according to https://status.odahub.io/ they are going up and down all the time, i.e. randomly refuse connections. But I did not actually see failing response personally. Might be some new network filter according to some origins.

> Seems I'm not authorized to merge here, so please do if you agree

Ok

volodymyrss:
> > These are definitely allowed by standards. Hope we won't encounter a need to use a path with url-encoded / some more allowed symbols
> 
> Would it be possible to provide an example/test ?

I think this comment was addressed, but please protest if not, @burnout87 !

burnout87:
> > > These are definitely allowed by standards. Hope we won't encounter a need to use a path with url-encoded / some more allowed symbols
> > 
> > 
> > Would it be possible to provide an example/test ?
> 
> I think this comment was addressed, but please protest if not, @burnout87 !

My bad, I forgot to request the change, in any case I guess we'll have examples in the future.

volodymyrss:
> > > > These are definitely allowed by standards. Hope we won't encounter a need to use a path with url-encoded / some more allowed symbols
> > > 
> > > 
> > > Would it be possible to provide an example/test ?
> > 
> > 
> > I think this comment was addressed, but please protest if not, @burnout87 !
> 
> My bad, I forgot to request the change, in any case I guess we'll have examples in the future.

Examples are included in the tests, or did you mean something else?

burnout87:
> 
> Examples are included in the tests, or did you mean something else?

it's ok, I didnt see the tests after they were added

PR #241, opened 2024-02-08
fix scw_list encoding

dsavchenko:
Will fix https://github.com/oda-hub/dispatcher-app/actions/runs/7828003199


dsavchenko:
Need to add CI to test dispatcher against changes in oda_api (as it's used internally there)

dsavchenko:
So dispatcher passes all test with this hotfix. Let's merge!

volodymyrss (approved):

PR #244, opened 2024-03-03
get_requested_resources method was added to Ontology

okolo:
Note, that I have temporarily changed the url of the ontology file to https://raw.githubusercontent.com/oda-hub/ontology/storage_resource_annotations/ontology.ttl in python-package.yml . This is needed for unit tests. As soon as [ontology](https://github.com/oda-hub/ontology/pull/23) pull request is done the change in python-package.yml should be reverted.


tests/test_ontology.py
```diff
@@ -342,3 +342,22 @@ def test_parsing_limits_bad_class(onto):
...
+
+def test_get_requested_resources(onto):
+    g = rdf.Graph()
+    annotated_ttl = add_prefixes + """
+            oda:notebook1 oda:usesRequiredResource oda:MyS3 .
+            oda:MyS3 a oda:S3 .
+            oda:MyS3 oda:resourceBindingEnvVarName "My_S3_CREDENTIALS" .
+            oda:notebook2 oda:usesOptionalResource oda:MyComputeResource .
+            oda:MyComputeResource a oda:ComputeResource .
+            oda:MyComputeResource oda:resourceBindingEnvVarName "MY_COMPUTE_CREDENTIALS" .
```
dsavchenko, line 355: This contradicts what you define [in ontology PR](https://github.com/oda-hub/ontology/blob/8c042f5219eeb43e7109989b234957110097e78c/ontology.ttl#L170-L171) that the domain is only a StorageResource. I think you can broaden the domain there

dsavchenko (approved):
Some concerns about performance due to creation of the additional graph (L450) and using a recursive function in the loop. (But I'm not really sure if the approach of complex SPARQL queries I use elsewhere is really better.)

Given that this method will be used once upon deployment, that's not a big problem.

tests/test_ontology.py
okolo: @dsavchenko , I have broaden the domain of resourceBindingEnvVarName property in ontology as you suggested

.github/workflows/python-package.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         python -m pip install types-simplejson types-requests types-click
         python -m pip install oda-knowledge-base[rdf,cwl]
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-        curl -o tests/oda-ontology.ttl https://raw.githubusercontent.com/oda-hub/ontology/main/ontology.ttl
+        curl -o tests/oda-ontology.ttl https://raw.githubusercontent.com/oda-hub/ontology/storage_resource_annotations/ontology.ttl
```
volodymyrss, line 34: I think we can merge the ontology now and adapt this.

okolo: Done

okolo:
The tests do not pass due to error in test_plot_tools_notebook, which has nothing to do with get_requested_resources

volodymyrss:
> The tests do not pass due to error in test_plot_tools_notebook, which has nothing to do with get_requested_resources

What is the error? If you notice it, could you make an issue? If it's complex to fix it quickly, please mark the tests skipped.

dsavchenko:
> What is the error? If you notice it, could you make an issue? If it's complex to fix it quickly, please mark the tests skipped.

This was some transient issue with the prod instance, I think. Fixed by rerunning. Could you merge, nb2workflow now depends on this change in master

volodymyrss (approved):

PR #266, opened 2024-07-09
Use deprecated node.js in "Executing notebook" workflow

dsavchenko:


.github/workflows/ci.yml
```diff
@@ -6,13 +6,16 @@ on:
...
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
+
 jobs:
   build:
     runs-on: ubuntu-latest
     container: integralsw/osa-python:auto-latest
     
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
```
burnout87, line 18: looks like this is not really necessary for this issue, but ok, it's good to upgrade if possible

burnout87 (approved):

.github/workflows/ci.yml
dsavchenko: Exactly

volodymyrss (approved):

PR #271, opened 2024-08-05
secret access API added based on renku secrets implementation

okolo:


volodymyrss:
Thanks! There is readthedocs in this repo, could you add a section there?

volodymyrss (changes requested):
Thanks! Please add a doc?

ferrigno:
Note that this PR is already published in docs at https://odahub.io/docs/guide-development/ Please merge it !

oda_api/secret.py
```diff
@@ -0,0 +1,11 @@
+import os
```
dsavchenko, line 1:
```suggestion
from typing import Optional
import os
```

oda_api/secret.py
```diff
@@ -0,0 +1,11 @@
...
+
+def get_secret(secret_name: str) -> str:
```
dsavchenko, line 3:
```suggestion
def get_secret(secret_name: str) -> Optional[str]:
```

dsavchenko (changes requested):
Static type checking fails, preventing unit tests from starting. Proposed changes should fix it.

dsavchenko:
Live API tests issue is most likely on the production instance side.

oda_api/secret.py
volodymyrss: @okolo , normally please have a look on this, but I will commit for now.

volodymyrss (changes requested):
Could you please fix the MyPy check?

PR #279, opened 2024-08-20
Remove dispatcher plot dependencies

ferrigno:
As discussed some time ago, I removed the dependency we had from Dispatcher by moving the class into plot_tools.
There is some code duplication from the two plot tools, but it looks reasonable.

@burnout87  please check if I was supposed to increase the version in setup.cfg and in case remove this change

I updated the requirements both n the requirements.txt file and in setup.py

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ matplotlib
 numpy
 jsonschema
 astroquery
--e git+https://github.com/oda-hub/dispatcher-app.git#egg=cdci_data_analysis[test]
 simplejson
 sentry_sdk
 rdflib
+bokeh
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.2.25
+current_version = 1.2.26
 commit = True
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+)(?P<build>\d+))?
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,8 @@ setup(name='oda_api',
             "astroquery",
             "scipy",
             "rdflib",
-            "black"
+            "black",
+            "bokeh"
         ],
       extras_require={
           'test': [
--- a/oda_api/plot_tools.py
+++ b/oda_api/plot_tools.py
@@ -30,6 +30,22 @@ import oda_api.api as api
 import time as _time
 import astropy.wcs as wcs
 
+from bokeh.layouts import row, gridplot
+#column,
+from bokeh.models import HoverTool
+#                          (CustomJS,
+#                           Toggle, 
+#                           RangeSlider,
+#                           HoverTool,
+#                           ColorBar,
+#                           LinearColorMapper,
+#                           LabelSet,
+#                           ColumnDataSource,
+#                           LogColorMapper)
+from bokeh.embed import components
+from bokeh.plotting import figure
+#from bokeh.palettes import Plasma256
+
 # NOTE GW, optional
 try:
     import ligo.skymap.plot
@@ -691,7 +707,7 @@ class OdaLightCurve(OdaProduct):
         return True
 
     def get_html_image(self, source_name, systematic_fraction, color='blue'):
-        import cdci_data_analysis.analysis.plot_tools
+        
         x, dx, y, dy, e_min, e_max = self.get_lc(source_name, systematic_fraction)
         mask = numpy.logical_not(numpy.isnan(y))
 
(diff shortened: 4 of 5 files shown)


requirements.txt
```diff
@@ -1,12 +1,15 @@
 requests
 future
-astropy
+astropy>=3.2
```
burnout87, line 3: maybe is it good to set it like in the dispatcher ?

https://github.com/oda-hub/dispatcher-app/blob/6e98f24012ab351d73a83801a42aafedfd3b202c/requirements.txt#L6

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.2.25
+current_version = 1.2.26
```
burnout87, line 2: this is not needed

burnout87 (changes requested):

requirements.txt
ferrigno: Done also in setup.py

setup.cfg
ferrigno: reverted back

requirements.txt
```diff
@@ -1,12 +1,15 @@
...
+pyjwt
 astroquery
--e git+https://github.com/oda-hub/dispatcher-app.git#egg=cdci_data_analysis[test]
-simplejson
-sentry_sdk
+scipy
 rdflib
+black
+bokeh
+-e git+https://github.com/oda-hub/dispatcher-app.git#egg=cdci_data_analysis[test]
```
burnout87, line 14: this has to be editable, otherwise tests will fail

dsavchenko:
> There is some code duplication from the two plot tools, but it looks reasonable.

That's ok, then we will be able to remove the duplication from the dispatcher code.

volodymyrss (approved):

setup.py
```diff
@@ -42,13 +42,15 @@
             "astroquery",
             "scipy",
             "rdflib",
-            "black"
+            "black",
+            "bokeh"
```
dsavchenko, line 46: In dispatcher, we fixed the bokeh version to be exactly `2.4.2`
The purpose of this is that it has to coincide with the version of the accompanying js library in the frontend. 

On the other hand, it's probably a good time to update it to the latest version, which is currently 3.5.1. One of the motivations is that, strictly speaking, the support for python 3.10 was added in bokeh 3.0, and we test oda_api package against 3.8 - 3.11

dsavchenko (changes requested):

setup.py
dsavchenko: Some complication is that it should be done in coordinated fashion between oda_api, dispatcher, frontend, and possibly gallery (not sure, please confirm)

dsavchenko: > same for the frontend

Frontend serves bokeh js libs that are used in conjunction with what dispatcher returns. That's why I also thought about gallery, if there are any bokeh-based products there. 
So the change is small, just to updating the version, but it's backwards-incompatible.

dsavchenko: > We can probably drop 3.8 and maybe 3.9

Well, 3.9 still has a year of lifecycle, but [Bokeh is officially supported and tested on Python 3.10 and above (CPython).](https://docs.bokeh.org/en/latest/docs/first_steps.html)
Still, I suspect it will work with it

volodymyrss: oda-api should run in different environments, sometimes pretty old ones. For legacy purposes. So we should keep as dependency as broad as possible.

burnout87: with https://github.com/oda-hub/oda_api/pull/279/commits/693e88deb99149dbe4f32dfd9baf204f9e244fb6 the html output for LCs will have bokeh-3.5.1

dsavchenko: > with [693e88d](https://github.com/oda-hub/oda_api/commit/693e88deb99149dbe4f32dfd9baf204f9e244fb6) the html output for LCs will have bokeh-3.5.1

Need to try, how it works with frontend, because it injects bokeh==2.4.2 in a page head

burnout87: > with [693e88d](https://github.com/oda-hub/oda_api/commit/693e88deb99149dbe4f32dfd9baf204f9e244fb6) the html output for LCs will have bokeh-3.5.1

I forgot to specify, the update is about the `get_html_image` function used to generate LC plots for the gallery.

For the frontend, a dedicated PR is needed on the dispatcher

dsavchenko: I didn't quite follow how it works in gallery. Each plot there have their own bokeh version injected, and they don't conflict because each product is on a different page, right? 
In frontend, it may be more complicated, say some products are cached with previous version of bokeh, and some other is generated with recent version injected. But all the tabs are on the same page, and here we may encounter problems

burnout87: > I didn't quite follow how it works in gallery. Each plot there have their own bokeh version injected, and they don't conflict because each product is on a different page, right? 

Exactly, more specifically, it is done here:
https://github.com/oda-hub/oda_api/blob/02e47e3670542e763f381364c4a87de99a2a9de9/oda_api/plot_tools.py#L715-L719

> In frontend, it may be more complicated, say some products are cached with previous version of bokeh, and some other is generated with recent version injected. But all the tabs are on the same page, and here we may encounter problems

Yes I can totally see it. And a potential workaround would require probably a lot of work. is it for sure not retro-compatible ?

dsavchenko: I saw failures when bokeh version in python and js are different. Between major releases, they surely won't be compatible.

volodymyrss (approved):

dsavchenko (approved):

burnout87 (approved):

PR #286, opened 2024-10-16
Test rmf

ferrigno:
This test deals with serialization of a variable size array in a response file.


burnout87:
Can be closed as the issue was solved with #287

PR #287, opened 2024-10-16
extra dump before decode

volodymyrss:
includes #286


oda_api/data_products.py
```diff
@@ -307,7 +312,7 @@ def to_fits_hdu(self):
             logger.debug('inside to_fits_hdu methods')
             logger.debug(f'name: {self.name}')
             logger.debug(f'header: {self.header}')
-            logger.debug(f'data: {self.data}')
+            # logger.debug(f'data: {self.data}')
```
volodymyrss, line 315:
```suggestion
            logger.debug(f'data: {self.data}')
```

oda_api/data_products.py
```diff
@@ -635,9 +643,11 @@ def from_fits_file(cls,filename,ext=None,hdu_name=None,meta_data={},name=''):
             for hdu in hdul:
                 if hdu.name == hdu_name:
                     _hdul.append(hdu)
+                # this ensures that the hdu is read, necessary for the case when hdu_name is not found
```
volodymyrss, line 646:
```suggestion
```

burnout87 (commented):
I guess now the tests can be removed/adapted?

if it's ok for you I can do it

volodymyrss:
It's important to keep the test, at least the one I added, to capture the problem and make sure it does not repeat with future updates.

The `test_rmf` can be updated as you prefer. 
Although in principle it is a higher-level test so it's good to keep it too.
It was important to show that the patch I provided is fixing the test @ferrigno added to demonstrate.
That is, it could be updated a bit if needed, much of it was not actually triggering for me.
What do you propose exactly?

burnout87:
>  
> What do you propose exactly?

Nothing in particular. Yes, to keep especially your test is important. As the `test_rmf`, also for me, I dont have the second exception caught too (I managed to have when running the same steps on a notebook...probably some differences in the environments used), so perhpas just that test can be adapted.

volodymyrss:
> > What do you propose exactly?
> 
> Nothing in particular. Yes, to keep especially your test is important. As the `test_rmf`, also for me, I dont have the second exception caught too (I managed to have when running the same steps on a notebook...probably some differences in the environments used), so perhpas just that test can be adapted.

I suppose there is a reason @ferrigno added this code, so if this motivation is written in a comment like I did, we could preserve it too.
If not useful let's reduce or drop it. Up to you and @ferrigno .

burnout87 (approved):

PR #290, opened 2024-10-18
adapt for py3.13, pyproject.yaml

burnout87:


dsavchenko:
What is the motivation for this change?

If doing this, maybe it is reasonable to migrate to `pyproject.toml`-only configuration?

burnout87:
> What is the motivation for this change?
> 

https://github.com/conda-forge/oda-api-feedstock/pull/28#issuecomment-2422440585

> If doing this, maybe it is reasonable to migrate to `pyproject.toml`-only configuration?
> 

I will look into it, shouldn't be too complicated

volodymyrss:
could you please refer to justification why we are doing this?

volodymyrss:
> What is the motivation for this change?
> 
> If doing this, maybe it is reasonable to migrate to `pyproject.toml`-only configuration?

This seems better than distutil.
@burnout87 could you please refer to current guidelines from python, pip?

burnout87:
> 
> This seems better than distutil. @burnout87 could you please refer to current guidelines from python, pip?

These seem to be https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

burnout87:
> could you please refer to justification why we are doing this?

I replied with a link tho the issue/PR where we talked about it, plus this PR is linked to an issue already

burnout87:
This explains the issue even better

https://conda-forge.org/news/2024/08/21/sunsetting-pip-deps/

volodymyrss (approved):
thanks!

PR #297, opened 2024-11-15
Support for optional parameters

dsavchenko:


oda_api/api.py
```diff
@@ -922,7 +924,9 @@ def _decode_res_json(self, res):
             else:
                 res = ast.literal_eval(str(res).replace('null', 'None'))
 
+            # what is it for?
             self.dig_list(res)
+
```
burnout87, line 929: it looks like it simply logs the content of the response...and report in case the content is not logga-ble...but I dont know why...

burnout87 (approved):

oda_api/api.py
dsavchenko: I wasn't even aware of that this comment ended up in PR...
It's doing it in a weird way, but in the end, if it works, don't touch it ;)

PR #322, opened 2025-02-04
Breakdown big gallery requests

burnout87:


ferrigno (approved):
I think that this is perfect for products that are contained in a single revolution, but what about products spanning the full mission ? How can we intercept them?

burnout87:
> I think that this is perfect for products that are contained in a single revolution, but what about products spanning the full mission ? How can we intercept them?

Very good point, thanks.
I will investigate if this can be achieved easily on the drupal side. If that's the case, then, the adaption here is going to be easy.

burnout87:
> I think that this is perfect for products that are contained in a single revolution, but what about products spanning the full mission ? How can we intercept them?

Ok I tried now to implement taking this into account, trying also, to optimize the calls

PR #334, opened 2025-05-26
patch live tests to use prod dispatcher only

dsavchenko:


dsavchenko:
> No this is different, it's a backend issue

Right, I just remember you also observed backend issues with requests coming from those notebooks before

dsavchenko:
And do you have any idea of what this backend exception means?

volodymyrss (approved):

Issue #340, opened 2025-12-09
Use integral time system in lightcurve plot


dsavchenko:
OdaLightcurve.build_fig use reference timestamp not accounting for 64,184 sec shift of Integral timesystem https://www.astro.unige.ch/integral//support/faq.cgi?DATA-007

PR #341, opened 2025-12-10
Correction of plot_tools for Acs time stamp and plotting of Polar LC

ferrigno:
This solves issue #340

### DIFF
--- a/tests/test_data_products.py
+++ b/tests/test_data_products.py
@@ -197,7 +197,7 @@ def test_bin_image():
     assert bin_image_decoded.binary_data == figdata
 
 
-@pytest.mark.parametrize('times,values,time_format,expected_units_dict_variants',
+@pytest.mark.parametrize('times,values,time_format,expected_units_dict_variants,timedel',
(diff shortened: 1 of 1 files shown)


oda_api/data_products.py
```diff
@@ -765,6 +765,7 @@ class LightCurveDataProduct(NumpyDataProduct):
     @classmethod
     def from_arrays(cls,
                     times, 
+                    timedel = None,
```
dsavchenko: This finally isn't used

oda_api/plot_tools.py
```diff
@@ -519,26 +532,39 @@ def get_lc(self, source_name, systematic_fraction=0):
         if self.instrument == 'SPI-ACS':
             self.timezero = hdu.header['TIMEZERO'] / 86400. + hdu.header['MJDREF']
             from astropy.time import Time
```
dsavchenko: Let's keep imports on top

oda_api/data_products.py
ferrigno: I have implemented it and modified the relative test.

